PR #1, opened 2016-06-02
Fixes the intro screens on 3.5" iphone

colinta:
Of course I tested this JUST after I submitted 1.8.0 :-(


steam:
LGTM, do we need this in 1.8.0 or are we going to wait?

caseydoran:
I'd like to wait since we're in review.

colinta:
Merging, but yeah this'll be in 1.8.x or 1.9.0

PR #2, opened 2016-06-02
Minor project group adjustment

orta:
### Before
screen shot 2016-06-02 at 6 09 51 pm

Sources was hidden midway, but was actually the first thing I was looking for

### After

screen shot 2016-06-02 at 6 11 49 pm

Now I can see it straight away, and know what I'm looking for

### DIFF
--- a/Ello.xcodeproj/project.pbxproj
+++ b/Ello.xcodeproj/project.pbxproj
@@ -2326,13 +2326,13 @@
 			path = Resources;
 			sourceTree = "<group>";
 		};
-		12F0E3AC1A4333370012CEA2 /* Sources */ = {
+		12F0E3AC1A4333370012CEA2 /* Ello */ = {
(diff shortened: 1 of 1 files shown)


steam:
Thanks for the PR @orta 👍. We have a convention of using `Sources` but I can see how it might be confusing. I struggle with group names differing from their folder. If we make this change I think we should consider changing the directory name from `Sources` to `Ello`. You have an opinion on this @colinta?

colinta:
I would say let's split the difference - move it to the top, but keep it as `Sources`.  We could re-order the other folders, too, by order of "relevance".

colinta:
I'm going to tackle this in another branch (see 4ff6fb5).  Good call, and I'm going to go even further by reordering MORE items and moving `Ello.entitlements` into `Support/`.

PR #4, opened 2016-06-03
Gits rid of git@ to ease cloning without keys.

valeriyvan:


steam:
Thanks for the PR @valeriyvan :metal:

PR #7, opened 2016-06-03
adds the robrix/Box license

colinta:


steam:
Giving this a thumbs up for now. We should try using the pod `pod 'Box', '~> 2.0'` moving forward. Do you remember why we didn't initially? My (terrible) memory is that we had a naming conflict at the time.

colinta:
Box doesn't have a podspec

steam:
How does it show up in the search results with a copyable pod file line?

colinta:
https://github.com/robrix/Box

¯\_(ツ)_/¯

colinta:
I guess someone else is maintaining the pod - I really don't know.

steam:
I've seen repos like this before that have no pod spec but have pods none-the-less. Black Magic 🌈 .

colinta:
> Maintained by Gordon Fontenot, Ash Furrow

https://cocoapods.org/pods/Box

steam:
Sure enough. "Sean, RTFM" is what you meant to say. 😏

PR #13, opened 2016-06-07
Adds the "flag user" feature

colinta:
- refactored `BlockUserModalViewController` to use a `Screen` instead of `.xib`
- added SnapKit!  Fun times.
- updated specs


Sources/Controllers/Relationship/BlockUserModalScreen.swift
```diff
@@ -0,0 +1,221 @@
...
+
+// MARK: ACTIONS
+
+    func blockTapped(sender: UIButton) {
+        let relationshipPriority: RelationshipPriority
+        if sender.selected == true {
+            relationshipPriority = .Inactive
+        } else {
+            relationshipPriority = .Block
+        }
```
steam: Any reason not to use `let relationshipPriority: RelationshipPriority = sender.selected ? .Inactive : .Block`?

colinta: Not at all, but the code below is *similar but a tiny bit different*, and I think breaking it out shows the difference more than if they're on one line.  In general I'm not a huge fan of the ternary op (not to say I never use it, though).

Sources/Controllers/Relationship/BlockUserModalViewController.swift
```diff
@@ -77,88 +52,58 @@ public class BlockUserModalViewController: BaseElloViewController {
...
-            blockButton?.selected = true
-        default: resetButtons()
+            closeModal() {
+                flagger.displayFlaggingSheet()
+            }
         }
     }
 
+    public func closeModal() {
+        closeModal {}
```
steam: What does this do?

Sources/Controllers/Stream/PostbarController.swift
```diff
@@ -290,7 +290,7 @@ public class PostbarController: NSObject, PostbarDelegate {
     }
 
     public func flagCommentButtonTapped(indexPath: NSIndexPath) {
-        if let comment = commentForIndexPath(indexPath) {
+        if let comment = commentForIndexPath(indexPath), presentingController = presentingController {
```
steam: Maybe drop this to the second line? Seems like a good standard to follow. What are you thoughts?

Sources/Controllers/Relationship/BlockUserModalViewController.swift
colinta: One version of that method accepts the `completion` param, but for conformance w/ the `Delegate` methods, I had to define a separate, parameter-less method (it wouldn't accept a method with a default argument of ` = {}`).  I forgot about this little hack, lemme look at that and see if it can be improved.

Sources/Controllers/Stream/PostbarController.swift
colinta: 👍

Specs/Controllers/Relationship/BlockUserModalViewControllerSpec.swift
```diff
@@ -61,153 +50,71 @@ class BlockUserModalViewControllerSpec: QuickSpec {
...
+                    let muteButton = (subviewThatMatches(subject.view) { ($0 as? UIButton)?.currentTitle == "Mute" }) as? UIButton
+                    let blockButton = (subviewThatMatches(subject.view) { ($0 as? UIButton)?.currentTitle == "Block" }) as? UIButton
+                    return (subject, muteButton, blockButton)
                 }
 
                 it("sets state properly when initialized with mute") {
-                    expect(subject.muteButton!.selected).to(beTrue())
-                    expect(subject.blockButton!.selected).to(beFalse())
+                    let (_, muteButton, blockButton) = blockButtons(.Mute)
+                    expect(muteButton?.selected).to(beTrue())
```
steam: We might want to convert these over to the `== true` syntax that we've been moving towards.

Specs/QuickExtensions.swift
```diff
@@ -214,3 +214,28 @@ public func haveImageRegion<S: OmnibarScreenProtocol>(equal image: UIImage) -> N
...
+public func subviewThatMatches(view: UIView, test: (UIView) -> Bool) -> UIView? {
+    for subview in allSubviews(view) {
+        if test(subview) {
+            return subview
+        }
+    }
+    return nil
+}
+
+public func haveSubview<V: UIView>(thatMatches test: (UIView) -> Bool) -> NonNilMatcherFunc<V> {
```
steam: Nice add :metal:

Sources/Controllers/Relationship/BlockUserModalScreen.swift
```diff
@@ -0,0 +1,221 @@
...
+
+        self.addSubview(backgroundButton)
+        self.addSubview(modalView)
+
+        let modalViews: [UIView] = [closeButton, titleLabel, muteButton, muteLabel, blockButton, blockLabel, flagButton, flagLabel]
+        for view in modalViews {
+            modalView.addSubview(view)
+        }
+
+        styleView()
```
steam: I like how you you've broken this up a bit. I wonder if we should have an additional function for the target/action connections rather than using `styleView()` for that.

Sources/Controllers/Relationship/BlockUserModalScreen.swift
```diff
@@ -0,0 +1,221 @@
...
+        button.titleLabel?.textColor = .whiteColor()
+    }
+
+    private func setText() {
+        muteButton.setTitle(InterfaceString.Relationship.MuteButton, forState: UIControlState.Normal)
+        blockButton.setTitle(InterfaceString.Relationship.BlockButton, forState: UIControlState.Normal)
+        flagButton.setTitle(InterfaceString.Relationship.FlagButton, forState: UIControlState.Normal)
+    }
+
+    private func arrangeViews() {
```
steam: An alternate approach we could take with `arrangeViews` is to move the layout code into a separate type. Something that is not a view. Perhaps a struct like `BlockUserModalLayout` that takes the views in it's initializer. Thoughts?

steam: That might allow us to slim down the `XYZScreen` files and pave the way for easy view model style specs in the future.

colinta: If this `Screen` needs to update & animate constraints based on view changes, that could complicate the `Screen/Layout` combo.  The two would, no matter what, be related 1:1.  Doesn't "feel" like a candidate for a struct, but I dunno!

Sources/Controllers/Relationship/BlockUserModalScreen.swift
```diff
@@ -0,0 +1,233 @@
...
+        bindActions()
+        setText()
+        arrange()
+    }
+
+    required public init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    public func setDetails(userAtName userAtName: String, relationshipPriority: RelationshipPriority) {
```
colinta: Note: move this to `init`, using a controller / screen configuration struct.

Sources/Controllers/Relationship/BlockUserModalViewController.swift
```diff
@@ -7,54 +7,25 @@
...
-    public var muteText: String {
-        return String(format: InterfaceString.Relationship.MuteWarningTemplate, userAtName, userAtName)
-    }
-
-    public var blockText: String {
-        return String(format: InterfaceString.Relationship.BlockWarningTemplate, userAtName)
-    }
+    var screen: BlockUserModalScreen!
 
     required public init(userId: String, userAtName: String, relationshipPriority: RelationshipPriority, changeClosure: RelationshipChangeClosure) {
```
colinta: Change to be a configuration struct

colinta:
(rebasing onto master)

PR #17, opened 2016-06-10
Adds support for `native_redirect` universal links

colinta:
These were being parsed incorrectly, and eventually leading to a call to `UIApplication.openURL`, hence the bouncing back and forth.

### DIFF
--- a/Sources/Networking/ElloURI.swift
+++ b/Sources/Networking/ElloURI.swift
@@ -184,6 +184,18 @@ public enum ElloURI: String {
         if path.beginsWith("ello://") {
             return path.stringByReplacingOccurrencesOfString("ello://", withString: "\(baseURL)/")
         }
+        if let url = NSURLComponents(string: path),
+            items = url.queryItems
+            where path.contains("native_redirect")
+        {
+            for item in items {
+                if let value = item.value
+                    where item.name == "target"
+                {
+                    return value
+                }
+            }
+        }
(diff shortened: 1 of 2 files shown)


steam:
Nice! We should talk with @jayzes about this as we can probably get rid of the `native_redirect` query string param all together now that we are going to go with universal links and no pre iOS 9 deep linking.

jayzes:
Totally - there's a bit of work on the server side to stop doing that, but not a lot. Would we be good to ship that change with 1.8 as it is then this becomes cleanup?

colinta:
1.8 shipped, do you mean 1.9?

I recommend pushing this out ASAP, either as 1.9 (which would include the ability to delete comments on a repost of your post) or as 1.8.1 (semantically this feels more correct, but whatev).

jayzes:
Sorry, words are difficult. I mean shipping the server-side change with 1.8 in the wild as-is, making it so we don't have to worry about `native_redirect` paths with > 1.8. Phrased differently, is there an issue if I change the server to just open those paths directly and ship it?

colinta:
Ahh, gotcha - changing the server side code (if that can happen quickly) would be ideal, since no release would be necessary.  I just don't want it to be next week, and this is still "in the wild".

steam:
Gonna close this since https://github.com/ello/ello/pull/1733 makes this irrelevant.

PR #19, opened 2016-06-10
Store the updated post AFTER posting the notification

colinta:
Otherwise the "still in memory" post will have its comment count updated twice


steam:
Nice @colinta, this one eluded me back when I looked into it months ago.

PR #21, opened 2016-06-14
refactor toggleSetting()

steam:
This was mostly an exercise in using `guard`, `map` and `filter` instead of `if let` and for loops. Do you think it is more readable @colinta?

### DIFF
--- a/Sources/Controllers/Settings/DynamicSettingCategoryViewController.swift
+++ b/Sources/Controllers/Settings/DynamicSettingCategoryViewController.swift
@@ -71,53 +71,52 @@ class DynamicSettingCategoryViewController: UIViewController, UITableViewDataSou
 }
 
 extension DynamicSettingCategoryViewController: DynamicSettingCellDelegate {
+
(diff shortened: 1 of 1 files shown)


Sources/Controllers/Settings/DynamicSettingCategoryViewController.swift
```diff
@@ -71,53 +71,52 @@ class DynamicSettingCategoryViewController: UIViewController, UITableViewDataSou
...
-                        value: currentUser.propertyForSettingsKey(setting.key),
-                        isVisible: DynamicSettingCellPresenter.isVisible(setting: setting, currentUser: currentUser)
-                    ))
-                }
-            }
+        guard let nav = self.navigationController as? ElloNavigationController,
+            currentUser = currentUser,
+            category = self.category else { return }
+
+        let visibility = category.settings.enumerate().map { (index, setting) in
```
steam: This is a cool way to get the index in a `map`. I'm not sure how efficient it is tho.

PR #24, opened 2016-06-17
Adds a few more Swiftlint rules

steam:
The following rules were added and fixed:

control_statement
legacy_constant
legacy_constructor
variable_name_max_length


colinta:
I dunno about *TINY*, but looks good! 😜

PR #26, opened 2016-06-22
Remove nib IBAction tap in embed cells

steam:
This fixes a regression introduced when we added double tap to love. We removed the IBAction in `StreamImageCell` but forgot to remove it in the subclass `StreamEmbedCell` nib.

[fixes #122085233](https://www.pivotaltracker.com/story/show/122085233)


colinta:
Thanks!

PR #30, opened 2016-06-24
Remove streamkind from preloadImage calls

steam:
`Preloader().preloadImages(...)` never used the `streamKind` parameter. Less code is the best code!


colinta:
Nice.

PR #31, opened 2016-06-24
Adds Categories to Discover

colinta:
- [x] Add images to `CategoryCardCell`

Technically still need to get the image URLs from Category, but that can be a tiny PR next week (API blocker).


Podfile
```diff
@@ -27,7 +27,7 @@ def ello_app_pods
   # swift pods
   pod 'TimeAgoInWords', git: 'https://github.com/ello/TimeAgoInWords'
   pod 'WebLinking', '~> 1.0'
-  pod 'SnapKit'
+  pod 'SnapKit', git: 'https://github.com/SnapKit/SnapKit'
```
steam: Do we want to point this at a specific sha?

Sources/Controllers/BaseElloViewController.swift
```diff
@@ -65,7 +65,10 @@ public class BaseElloViewController: UIViewController, ControllerThatMightHaveTh
 // MARK: Search
 public extension BaseElloViewController {
     func addSearchButton() {
-        elloNavigationItem.rightBarButtonItem = UIBarButtonItem(image: InterfaceImage.Search.normalImage, style: .Done, target: self, action: #selector(BaseElloViewController.searchButtonTapped))
+        elloNavigationItem.rightBarButtonItem = UIBarButtonItem(
+            image: InterfaceImage.Search.normalImage, style: .Done,
```
steam: I like splitting long lines into multiple ones but I'd vote for putting each argument on it's own line when we do. Thoughts?

Sources/Controllers/Discover/DiscoverAllCategoriesScreen.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  DiscoverAllCategoriesScreen.swift
+//  Ello
+//
+//  Created by Colin Gray on 6/17/2016.
+//  Copyright (c) 2016 Ello. All rights reserved.
+//
+
+public class DiscoverAllCategoriesScreen: ElloScreen {
```
steam: Is this suppose to be empty?

Sources/Controllers/Discover/DiscoverAllCategoriesViewController.swift
```diff
@@ -0,0 +1,78 @@
+//
+//  DiscoverAllCategoriesViewController.swift
+//  Ello
+//
+//  Created by Colin Gray on 6/17/2016.
+//  Copyright (c) 2016 Ello. All rights reserved.
+//
+
+public class DiscoverAllCategoriesViewController: StreamableViewController {
+    var screen: DiscoverMoreCategoriesScreen!
```
steam: Whats the thinking about waiting until `loadView()` to create this rather than creating it in `init()` and bypassing the implicitly unwrapped optional, performance optimization?

colinta: Just trying to be a good citizen - since "view loading" is the same as "screen loading", I thought it best to create the screen in `loadView`.  I'm not stuck on this, it's just what seemed natural.  TBH, I don't understand the separation of `controller / view`, I'd prefer that the view just *always be there*.

Sources/Controllers/Discover/DiscoverAllCategoriesScreen.swift
colinta: Yeah in case we need to add things later, it's a placeholder, essentially.

Sources/Controllers/Discover/DiscoverAllCategoriesViewController.swift
```diff
@@ -0,0 +1,78 @@
...
+// MARK: StreamViewDelegate
+extension DiscoverAllCategoriesViewController {
+    override public func streamViewStreamCellItems(jsonables: [JSONAble], defaultGenerator generator: StreamCellItemGenerator) -> [StreamCellItem]? {
+        var items: [StreamCellItem] = []
+        if let categories = jsonables as? [Category] {
+            let sortedCategories = categories
+                .sort { $0.order < $1.order }
+            for category in sortedCategories {
+                items.append(StreamCellItem(jsonable: category, type: .CategoryCard))
+            }
```
steam: This could also be:
```
items = categories.sort { 
    return $0.order < $1.order 
}.map { 
    return StreamCellItem(jsonable: $0, type: .CategoryCard) 
}
```

steam: I hesitated before writing that ☝️ as I didn't want to annoy you. Just a discussion really. Not intending to bike shed. :)

Sources/Controllers/Discover/DiscoverViewController.swift
```diff
@@ -7,43 +7,88 @@
...
+
+        sharedInit()
+        streamViewController.streamKind = .Discover(type: .Featured)
+    }
+
+    private func sharedInit(title title: String = InterfaceString.Discover.Title) {
+        self.title = title
+        elloNavigationItem.title = title
+
+        if title != InterfaceString.Discover.Title {
```
steam: Is this a check to see if we've navigated to more detailed category?

Sources/Controllers/Discover/DiscoverAllCategoriesViewController.swift
colinta: No I like it!  Pushing it up.

Sources/Controllers/Discover/DiscoverViewController.swift
colinta: It is, yeah - a boolean var would be more explicit & useful here.

Podfile
colinta: We can point to 0.30 now that it's released, will update.

Sources/Controllers/BaseElloViewController.swift
colinta: I don't have a strong opinion - as long as its readable, it's fine w/ me.  I left `.Done` because it doesn't matter what the style is when using an image (afaik).

Sources/Controllers/Discover/DiscoverAllCategoriesViewController.swift
colinta: Another option here:

```swift
// var screen: DiscoverScreen!
var screen: DiscoverScreen { return self.view as! DiscoverScreen }
```
This would have the same side effects as calling `self.view` (`loadView` gets called), and keep the code terse & easy to reason about.

Sources/Controllers/Profile/ProfileViewButtons.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  ProfileViewButtons.swift
```
steam: Nice refactor 👍

Sources/Controllers/Stream/CellDequeing/CategoryCardCellPresenter.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  CategoryCardCellPresenter.swift
```
steam: I've been deleting the comments at the top of files lately. I think git history is more useful, you down to get rid of them when you see them and not add them to new files? Maybe they serve a purpose?

Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -29,11 +29,17 @@ public protocol StreamEditingDelegate: class {
     func cellLongPressed(cell: UICollectionViewCell)
 }
 
-@objc
-public protocol StreamScrollDelegate: class {
+public typealias StreamCellItemGenerator = () -> [StreamCellItem]
+public protocol StreamViewDelegate: class {
```
colinta: @steam Definitely take a good look at this protocol - lots added, and they add hooks that basically *return nil for default behavior*

Sources/Controllers/Stream/CellDequeing/CategoryCardCellPresenter.swift
```diff
@@ -0,0 +1,25 @@
...
+public struct CategoryCardCellPresenter {
+
+    static func configure(
+        cell: UICollectionViewCell,
+        streamCellItem: StreamCellItem,
+        streamKind: StreamKind,
+        indexPath: NSIndexPath,
+        currentUser: User?)
+    {
+        if let cell = cell as? CategoryCardCell,
```
steam: Question for us here. When do we use `guard`? Here is my thought, if the entire function is basically an `if let` around a chunk of code we should use `guard` and pull the body of the function up a indentation level. Thoughts?

colinta: For sure - follow up question: do we update all the presenters?

steam: Yeah, thats the tricky part. As swift has evolved how do we incrementally evolve our code with it? If we only update the thing we're working on the PR stays focused on the feature but the code is slightly different in similar but unrelated places. I don't know the best answer. I suppose we do when we think its not a huge effort?

Sources/Controllers/Stream/CellDequeing/CategoryCardCellPresenter.swift
colinta: Works for me, I have to take additional steps to *add* these copyright blocks in there, but seems like `LICENSE.txt` should cover all these files.

Sources/Controllers/Stream/Cells/StreamCellType.swift
```diff
@@ -201,10 +246,14 @@ public enum StreamCellType: Equatable {
 
     public var oneColumnHeight: CGFloat {
         switch self {
+        case Category, SeeAllCategories:
```
steam: Should this be two separate lines?

Sources/Controllers/Stream/StreamDataSource.swift
```diff
@@ -238,82 +243,77 @@ public class StreamDataSource: NSObject, UICollectionViewDataSource {
...
-            let streamCellItem = visibleCellItems[indexPath.item]
-
-            let cell = collectionView.dequeueReusableCellWithReuseIdentifier(streamCellItem.type.name, forIndexPath: indexPath)
-
-            return cell
-
-        }
-        return UICollectionViewCell()
+        guard indexPath.item < visibleCellItems.count else {
+            return UICollectionViewCell()
```
steam: Nice use of guard 👍

Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -765,6 +812,35 @@ extension StreamViewController {
...
+        vc.currentUser = currentUser
+        navigationController?.pushViewController(vc, animated: true)
+    }
+}
+
+// MARK: StreamViewController: CategoryDelegate
+extension StreamViewController: CategoryDelegate {
+
+    public func categoryCellTapped(cell: UICollectionViewCell) {
+        if let indexPath = collectionView.indexPathForCell(cell),
```
steam: Use `guard` here?

Sources/Controllers/Stream/Cells/StreamCellType.swift
colinta: It could be, of course, but this file has a lot of one-line-case statements.  Like some of the other refactors, I wonder again if we should change this in piece meal (and have inconsistencies), or do a follow up PR to edit it all at once.

Sources/Controllers/Discover/DiscoverAllCategoriesScreen.swift
colinta: I noticed later that this Screen was not being used in `DiscoverAllCategoriesViewController`, that's been fixed!

Sources/Model/Category.swift
```diff
@@ -0,0 +1,76 @@
...
+//  Created by Colin Gray on 6/14/2016.
+//  Copyright (c) 2016 Ello. All rights reserved.
+//
+
+import SwiftyJSON
+
+public let CategoryVersion = 1
+
+public final class Category: JSONAble, Groupable {
+    static let featured = Category(id: "meta1", name: "Featured", slug: "recommended", order: 0, level: .Meta)
```
steam: Should these names be in `InterfaceString`?

Sources/Model/CategoryLevel.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  CategoryLevel.swift
+//  Ello
+//
+//  Created by Colin Gray on 6/14/2016.
+//  Copyright (c) 2016 Ello. All rights reserved.
+//
+
+public enum CategoryLevel: String, Equatable {
+    case Meta = "meta"
```
steam: Do these need to have lower case string values? If not we get them for free if the `enum` is a `String`.

colinta: No they're defined by the server - we could make the enums lower case (have you noticed that Apple is using lower case enum values?  I don't like it, but that's what I saw in a lot of the videos).

Sources/Model/CategoryList.swift
```diff
@@ -0,0 +1,77 @@
...
+        }
+        return CategoryList(categories: categories)
+    }
+
+}
+
+extension CategoryList {
+    class func hardCodedPrimaries() -> CategoryList {
+        return CategoryList(categories: [
+            Category(id: "meta1", name: "Featured", slug: "recommended", order: 0, level: .Meta),
```
steam: Should these reference `InterfaceString` as well?

Sources/Model/CategoryLevel.swift
steam: I did see that. We should probably switch to lowercase. I agree that it is strange and don't particularly like it.

Sources/Model/Post.swift
```diff
@@ -54,6 +58,18 @@ public final class Post: JSONAble, Authorable {
...
+            return []
+        }
+        return categories
+    }
+    public var category: Category? {
+        guard let category = categories.first else {
+            return nil
+        }
+        return category
+    }
```
steam: Any reason not to change this?

```
public var category: Category? {
    return categories.first
}
```

Sources/Networking/Mapper.swift
```diff
@@ -29,17 +28,12 @@ public struct Mapper {
...
-            }
-            return jsonables
+    public static func mapToObjectArray(dicts: [[String:AnyObject]], fromJSON: FromJSONClosure) -> [JSONAble] {
+        let jsonables: [JSONAble] = dicts.map {
+            let jsonable = fromJSON(data: $0, fromLinked: false)
+            return jsonable
         }
-
-        return nil
+        return jsonables
```
steam: This could be simplified even more now I think. Something like:

```
public static func mapToObjectArray(dicts: [[String:AnyObject]], fromJSON: FromJSONClosure) -> [JSONAble] {
    return dicts.map { fromJSON(data: $0, fromLinked: false) }
}
```

Sources/Model/Post.swift
colinta: 💯

Sources/Networking/Mapper.swift
colinta: I'll do one better!

```swift
    public static func mapToObjectArray(dicts: [[String:AnyObject]], fromJSON: FromJSONClosure) -> [JSONAble] {
        return dicts.map { fromJSON(data: $0, fromLinked: false) }
    }

    public static func mapToObject(object: AnyObject?, fromJSON: FromJSONClosure) -> JSONAble? {
        return (object as? [String:AnyObject]).flatMap { fromJSON(data: $0, fromLinked: false) }
    }
```

steam:
Nice work @colinta 💪. Is this still WIP? LGTM.

Sources/Networking/Mapper.swift
steam: 🌹

colinta:
Cool!  The only last thing is/was the category image assets, but that's blocked on API, and can be merged later.

PR #32, opened 2016-06-24
Load Profile and Post Detail in small pieces

steam:
In an effort to speed up stream loading we are implementing generators that handle networking interactions (via services) and `StreamCellItem` management. Rather than loading everything a screen needs to be complete before rendering the screen we will load components independently and render them as they come in if possible.

The basic idea is that each `StreamableViewController` will have it's own concrete `StreamGenerator` that will be responsible for making 1 or more network requests (via the service layer) to asynchronously assemble a given stream's content. 

`StreamableViewController` subclasses will create a concrete `StreamGenerator`. `StreamableViewController` subclasses implement `StreamDestination` awaiting instruction from their custom `StreamGenerator`.

She be done!

- [x] Handle Page Loading Token
- [X] Allow updating of placeholder cells
- [x] Updating of groups of cells bound by `PlaceholderType`
- [x] Load primary `JSONAble` from Yap in `StreamableViewController` subclasses (`PostDetailViewController` and `ProfileViewController`)
- [x] Find commonalities and abstract them into the protocol 
- [x] Fix broken specs
- [x] Add generator specs
- [x] Verify paging works in `PostDetailViewController`
- [x] Verify paging works in `ProfileViewController`
- [x] Verify no results continues to work as expected
- [x] The profile header is a bit too short (links is 1/2 cut off).
- [x] Rewire list/grid toggle functionality in profile detail.
- [X] Ensure profile or post loads first (don't show other items until the main JSONAble is displayed) Marking this one done since we load the user/post from YAP prior to creating the generator. We'll need to keep an eye out for scenarios where YAP didn't have the post/user. Seems rare.
- [x] Do not display secondary responses if the primary request fails
- [x] Do not display secondary responses until the primary response is displayed


Sources/Controllers/Profile/ProfileViewController.swift
```diff
@@ -334,103 +318,48 @@ public class ProfileViewController: StreamableViewController {
...
-            streamViewController.appendStreamCellItems(headerItems)
-        }
-        userLoaded(user, responseConfig: responseConfig, isReload: false)
+//        guard let user = user, responseConfig = streamViewController.responseConfig else { return }
+//
+//        if let headerItems = headerItems {
+//            streamViewController.appendStreamCellItems(headerItems)
+//        }
+        // TODO: re-layout posts
+//        userLoaded(user, responseConfig: responseConfig, isReload: false)
```
steam: @colinta `userLoaded` was removed. I had a merge conflict with master here. There is a good chance I blew away the changes you made to show animated gifs in a header. Before we merge this PR we'll need to make sure animated avatars and gifs continue to work. Look down in the `StreamDestination` functions. We may need to update them.

Sources/Controllers/Stream/Cells/StreamCellType.swift
```diff
@@ -99,43 +104,8 @@ public enum StreamCellType: Equatable {
...
-        case UserAvatars: return 20
-        case UserListItem: return 21
-        case ColumnToggle: return 22
-        case Category: return 23
-        case CategoryCard: return 24
-        case SeeAllCategories: return 25
-        case CategoryList: return 26
-        }
+    public var identifier: String {
+        return "\(self)"
```
steam: This "works" but is not all that great. 'self' is often "Type(nil)" or similar. We may want something a bit more controlled.

Sources/Controllers/Stream/Cells/StreamTextCellSizeCalculator.swift
```diff
@@ -27,10 +27,28 @@ public class StreamTextCellSizeCalculator: NSObject, UIWebViewDelegate {
...
+        let job: CellJob = (cellItems: cellItems, width: width, columnCount: columnCount, completion: completion)
+        cellJobs.append(job)
+        if cellJobs.count == 1 {
+            processJob(job)
+        }
+    }
+
+// MARK: Private
+
+    private func processJob(job: CellJob) {
```
steam: The `processCells` and `processJob` functions seem so similar across calculators. I wonder if there is an opportunity to consolidate this. Thoughts?

colinta: Ideally, yes - but I wanted to get this working and out of the way before I delved into *yet another* refactor.

steam: Makes sense. I added #48 to remind us to take a look at some point.

Sources/Controllers/Stream/StreamDataSource.swift
```diff
@@ -224,23 +224,17 @@ public class StreamDataSource: NSObject, UICollectionViewDataSource {
...
-            temporarilyUnfilter() {
-                if let index = self.streamCellItems.indexOf(itemToRemove) {
-                    self.streamCellItems.removeAtIndex(index)
-                }
-            }
-        }
-    }
-
-    public func removeItemAtIndexPaths(indexPaths: [NSIndexPath]) {
+    public func removeItemsAtIndexPaths(indexPaths: [NSIndexPath]) {
```
steam: Good catch! Simplify :)

steam:
I spiked a bit on moving `StreamViewController` over to using a generator but it was too big. Too many things depend on it. We'll want to move it over eventually but lets keep this focused and get it 🚢 :metal:

steam:
@colinta, are you happy now? :)

colinta:
I'M NEVER HAPPY

steam:
Made the change 👍

colinta:
Nice!  That file is lookin' good 😃

steam:
@colinta feel free to merge this when you're ready for it

PR #35, opened 2016-06-28
These 'hardCodedPrimaries' are not used any more

colinta:
Separate PR just to make sure specs are still green.


colinta:
Self-reviewing, just waiting for specs to go 🍏

PR #37, opened 2016-07-05
Adds Category image and 3 columns category card display on ipad

colinta:


ShareExtension/ShareTrackerOverrides.swift
```diff
@@ -28,4 +28,8 @@ public class Tracker {
     func contentFlaggingCanceled(type: ContentType, contentId: String) {}
     func createdAtCrash(identifier: String, json: String?) {}
     func encounteredNetworkError(path: String, error: NSError, statusCode: Int?) {}
-}
\ No newline at end of file
+}
+
+public class Window {
```
colinta: This is accessed from `StreamKind`, so we need to "stub out" this window-size code.

PR #39, opened 2016-07-06
Show the post category in grid view

colinta:
Added specs for this in `StreamHeaderCellPresenterSpec`


Sources/Controllers/Stream/CellDequeing/StreamHeaderCellPresenter.swift
```diff
@@ -79,8 +79,7 @@ public struct StreamHeaderCellPresenter {
                 followButtonVisible = false
             }
 
-            let category: Category? = streamKind.isGridView ? nil : post?.category
```
colinta: This was the crux of the issue - the cell has control over whether the category is visible, not the presenter.

PR #49, opened 2016-07-11
Changes screen tracking

colinta:
1. Remove duplicate "Profile/Post viewed" events (the `screenAppeared` tracking takes care of these)
2. Override the `tracking` in `SimpleStreamViewController`

These should make life easier for @Justin-Holmes


Justin-Holmes:
:metal:

steam:
More red than green. :metal:

PR #50, opened 2016-07-11
Updates Repost decoration, and Snapshots

colinta:
Long ago I used `setNimbleTolerance(1)` to decrease false negative snapshot failures... turns out the value of `1` means "accept all differences" (not 1%, as I thought).

But we're good, only `CategoryListCell` had any issues (needed an extra `setNeedsLayout`), the other snapshots just needed to be updated.

Removed `RepostHeader`, the reposter is now part of the `StreamHeaderCell`, even in Grid view (per Lucian).


Specs/Controllers/Stream/StreamDataSourceSpec.swift
```diff
@@ -227,7 +227,7 @@ class StreamDataSourceSpec: QuickSpec {
                         it("returns the correct number of rows") {
                             // there should be 10 reposts
                             // 10 * 5(number of cells for a repost w/ 2 regions) = 50
-                            expect(subject.collectionView(vc.collectionView, numberOfItemsInSection: 0)) == 50
+                            expect(subject.collectionView(vc.collectionView, numberOfItemsInSection: 0)) == 40
```
colinta: Oops, I should update those comments (reposts have 4 cells instead of 5)

steam:
LGTM. DELETE IT ALL.

colinta:
```
Executed 1999 tests, with 0 failures (0 unexpected) in 101.157 (102.077) seconds
```

Damn snapshots - failing on Travis. :-|

colinta:
Almost there!

colinta:
Green!

PR #53, opened 2016-07-12
Adds Category support to deep linking

colinta:
:metal: @rynbjyn


rynbyjn:
Cool! One question, does this handle `/discover/featured`, `/discover/recommended`, `/discover`, and `/` as the same path? This is the regex we use in the webapp to determine this: `/^\/(?:discover(\/featured|\/recommended)?)?$/`

colinta:
I'll check - probably not, and probably easy to add support for those.

colinta:
Does now!

Supports: (w/ or w/out trailing slash)

```
/discover
/discover/featured
/discover/recommended
```

Sources/Networking/ElloURI.swift
```diff
@@ -11,11 +11,12 @@ import Keys
 
 public enum ElloURI: String {
     // matching stream or page in app
-    case Discover = "discover"
+    case Discover = "discover(/featured|/recommended)?/?$"
```
rynbyjn: I guess y'all don't need to support the logged out `/` being a discover stream too.

colinta: Yeah not at the moment

colinta: Or... do we?  Not sure where `"/"` would go, I'll check.

colinta: It does go to discover 👍

rynbyjn: tight :metal:

rynbyjn:
LGTM :metal:

colinta:
These tests are just lying, I dunno what's up: ```Executed 2001 tests, with 0 failures (0 unexpected) in 99.713 (100.641) seconds```

steam:
👍

Issue #55, opened 2016-07-13
ElloKeys has no member teamId


AndrewSB:
Doesn't compile on master: 
<img width="720" alt="screen shot 2016-07-13 at 11 10 50 am" src="https://cloud.githubusercontent.com/assets/3814772/16814401/89154f60-48ea-11e6-8659-ddbcbc16c30f.png">

colinta:
I'll check this out.

colinta:
Can you try adding this to the bottom of Podfile?

```ruby
plugin 'cocoapods-keys', {
  project: 'Ello',
  keys: [
    'ClientKey',
    'ClientSecret',
    'CrashlyticsKey',
    'Domain',
    'HttpProtocol',
    'Salt',
    'SegmentKey',
    'TeamId',   # <--- add this
  ]
}
```

colinta:
And then run `pod install`

AndrewSB:
Having a weird error inside the `ello-ios` directory - I'm being told that `pod not found`, but only inside the `ello-ios`.

```
➜  ~ which pod
> /Users/asb/.rvm/gems/ruby-2.3.0/bin/pod
➜  Developer cd Developer 
➜  Developer which pod
> /Users/asb/.rvm/gems/ruby-2.3.0/bin/pod
➜  Developer cd ello-ios 
➜  ello-ios git:(master) ✗ which pod
> pod not found
```

colinta:
That must have something to do with `.ruby-version` being set to 2.2.2. @666 can we remove that file?

rynbyjn:
I think you need to run the `pod` commands with `bundle exec` due to cocoapods being in the `Gemfile`.

AndrewSB:
yup, that worked @rynbyjn!

Maybe add an explanation to the Gemfile? It's probably just me and my lack of a ruby background

AndrewSB:
@colinta no luck.

When I `bundle exec pod install`, the CLI prompts me to enter in a key for `TeamId`, but it didn't affect ElloKeys - still no `teamId`

colinta:
OK, try this in the terminal:

```
bundle exec pod keys set ClientKey SHOULDNOTMATTER Ello
bundle exec pod keys set ClientSecret SHOULDNOTMATTER Ello
bundle exec pod keys set TeamId SHOULDNOTMATTER Ello
bundle exec pod keys set Domain SHOULDNOTMATTER Ello
[...]

PR #56, opened 2016-07-13
use didFinishAnimating to get new current page for onboarding pageVC

AndrewSB:
Noticed a UI bug in on boarding, current if you let go of the page before it finishes animating, the pageControl doesn't update.

I've found that using `didFinishAnimating` has fixed that in the past.

This hasn't been tested.

### DIFF
--- a/Sources/Controllers/Intro/IntroViewController.swift
+++ b/Sources/Controllers/Intro/IntroViewController.swift
@@ -105,8 +105,6 @@ public class IntroViewController: UIViewController, UIPageViewControllerDataSour
     {
         var index = (viewController as! IntroPageController).pageIndex!
 
-        pageControl.currentPage = index
-
         if index <= 0 {
             return nil
         }
@@ -120,8 +118,6 @@ public class IntroViewController: UIViewController, UIPageViewControllerDataSour
     {
         var index = (viewController as! IntroPageController).pageIndex!
 
-        pageControl.currentPage = index
-
         index += 1
 
         if index >= viewControllers.count {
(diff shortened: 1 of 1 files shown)


colinta:
Tested locally, and the page control isn't updated at all - it always indicates "page one". 😟

colinta:
The linter is also giving "trailing whitespace" violations.  These can be fixed by turning on the "Automatically trim trailing whitespace" option in XCode:

screen shot 2016-07-13 at 1 23 31 pm

colinta:
This method isn't being called:

```swift
public func pageViewController(pageViewController: UIPageViewController, didFinishAnimating finished: Bool, previousViewControllers: [UIViewController], transitionCompleted completed: Bool)
```

colinta:
I checked the delegate method name, and it you have it correct, so I'm not sure what's going on here!  I'll leave this open for now, in case you want to play with it more.

I'm also not able to reproduce the error - can you describe what you did to cause the error?

colinta:
Ah, I was able to reproduce.  Good eye!  It's hard to catch.

AndrewSB:
I'm still having problems building (related to #55), once I have it building I'll play around and see why the delegate method isn't being called 😄

colinta:
Ahh I see what it is!

```diff
diff --git a/Sources/Controllers/Intro/IntroViewController.swift b/Sources/Controllers/Intro/IntroViewController.swift
index a4e4e10..19c8c1a 100644
--- a/Sources/Controllers/Intro/IntroViewController.swift
+++ b/Sources/Controllers/Intro/IntroViewController.swift
@@ -25,6 +25,7 @@ public class IntroViewController: UIViewController, UIPageViewControllerDataSour

         pageViewController?.view.frame = frame
         pageViewController?.dataSource = self
+        pageViewController?.delegate = self

         // Load and set views/pages
         let welcomePageViewController = storyboard
```

We never assigned the delegate!

colinta:
Yup, that does the trick, your fix is working great now.  I'll add this commit before merging.

AndrewSB:
awesome! good catch @colinta

colinta:
```
Executed 2001 tests, with 0 failures (0 unexpected) in 101.530 (102.451) seconds
```

CI is likely going to flake out, so I'm merging now.

colinta:
Just wanted to let you know that this, and the changes to `Keyboard`, are going to be included in the 1.9 release of the Ello iOS app.  Big thanks!

AndrewSB:
Awesome!

Super exciting, happy I could help out 🙃

PR #58, opened 2016-07-13
Updates the ruby version to the latest mothership.

rynbyjn:


colinta:
Why do we even need this file?

rynbyjn:
It allows `rvm` or `rbenv` to choose the ruby version you want to use automatically in case there are issues with other versions of ruby and other dependencies in the project.

colinta:
I've never needed to use rvm or rbenv, so this file doesn't help or hinder me - I'm just wondering if it's "one more thing" that we need to keep up to date, even though it doesn't provide benefit.  What is in danger of breaking if we remove it?

rynbyjn:
Cool with me to remove it. Would only be gems that rely on a specific version of ruby that could be effected in the future.

colinta:
I think it's causing trouble w/ #55, so let's do that, and bring it back if it's needed.

PR #64, opened 2016-07-14
Update the onboarding logic

colinta:
See specs for a complete table of expectations, but essentially:

- only show onboarding once, on any platform
- if web value is nil, assume it's been seen
- if web and local are both < current, show onboarding

then:

- update the API as needed


colinta:
@jayzes 9452d11 has the specs

steam:
LGTM

PR #66, opened 2016-07-15
Adds support for affiliate image regions

colinta:
Looks for regions of type `affiliate_image`, and passes those to `ImageRegion` and `StreamImageCell`.

If `link_url` is present (it's assigned to `ImageRegion.affiliateURL`), the `affiliateButton` is displayed.


Sources/Controllers/Stream/Cells/StreamImageCell.swift
```diff
@@ -190,6 +232,12 @@ public class StreamImageCell: StreamRegionableCell {
         streamImageCellDelegate?.imageTapped(self.imageView, cell: self)
     }
 
+    @IBAction func affiliateTapped() {
+        if let affiliateURL = affiliateURL {
```
steam: Should this be a `guard`?

steam:
👍

colinta:
```Executed 2022 tests, with 0 failures (0 unexpected) in 141.647 (143.194) seconds```

PR #68, opened 2016-07-15
Fixes insertion point in Profile when creating a post

colinta:
We added 3 cells, so change `1` to `4`.

Magic numbers == 😢


jayzes:
✨

steam:
Lets hold off on this. We have the generators coming and this issue should be handled by the generators.

colinta:
OTOH, we could include this in 1.9.1, and then override it with the stream generators in 1.10.

steam:
I think we should since we're not including generators in the next build.

PR #70, opened 2016-07-18
Fixes to category button frame, and refactor repostedByButton frame

colinta:
No more unnecessary truncation on `categoryButton`, and the `repostedByButton` frame calculation was more confusing than necessary.


steam:
Should we have snapshot specs here to verify this?

colinta:
Good call, I'll update the existing ones to use a long category name.

steam:
:metal:

colinta:
Oops, just noticed another edge case (in grid, the category can/should be `26pt` wider). Another PR, built on top of this one.

PR #71, opened 2016-07-18
Attach a custom tracking event when a post is created via the ShareExtension

colinta:
I think the number of lines added and removed is approximately equal to how many months ago the story was created.

*Progress!*

### DIFF
--- a/ShareExtension/ElloAPISharingHeaders.swift
+++ b/ShareExtension/ElloAPISharingHeaders.swift
@@ -1,0 +1,14 @@
+////
+///  ElloAPISharingHeaders.swift
+//
+
+extension ElloAPI {
+    var sharingHeaders: [String: String]? {
+        switch self {
+        case .CreatePost:
+            return ["X-Segment-Event": "share-event"]
+        default:
+            return nil
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


steam:
:metal:

Sources/Networking/ElloAPISharingHeaders.swift
```diff
@@ -0,0 +1,7 @@
+////
+///  ElloAPISharingHeaders.swift
+//
+
+extension ElloAPI {
+    var sharingHeaders: [String: String]? { return nil }
```
colinta: No-op in the app, so all good there.

Sources/Networking/ElloAPI.swift
```diff
@@ -472,6 +472,10 @@ extension ElloAPI: Moya.TargetType {
             ]
         }
 
+        if let sharingHeaders = self.sharingHeaders {
+            assigned += sharingHeaders
```
colinta: Found headers, add them to the request.

ShareExtension/ElloAPISharingHeaders.swift
```diff
@@ -0,0 +1,14 @@
+////
+///  ElloAPISharingHeaders.swift
+//
+
+extension ElloAPI {
+    var sharingHeaders: [String: String]? {
+        switch self {
+        case .CreatePost:
+            return ["X-Segment-Event": "share-event"]
```
colinta: The `X-Segment-Event` header is *only* added for the `CreatePost` endpoint.

PR #73, opened 2016-07-18
Only show category in .Discover(.Featured) stream

colinta:
Last PR for 1.9.1 - hide category everywhere *except* the `discover/featured/` stream


steam:
👍

PR #74, opened 2016-07-19
Updates the headers in Specs/ folder

colinta:
I must have excluded this from my regex earlier


steam:
Yowzas. Thats a lot of files. 👍

PR #81, opened 2016-07-25
Resolves #79

### LINKED ISSUE #79: AuthToken.reset()
Slightly confused on how `AuthToken's` reset function works:

```
static func reset() {
        var keychain = sharedKeychain
        keychain.authToken = nil
        keychain.refreshAuthToken = nil
        keychain.authTokenType = nil
        keychain.isPasswordBased = false
        keychain.username = nil
        keychain.password = nil
    }
```

If you look at the setters for the properties of `ElloKeychain`, they all perform checks for and prevent nil values from being set. Obviously you can nil a keychain value, but I'm failing to see how the design of `ElloKeychain` allows for this.

colinta:
Refactors `ElloKeychain.swift` to allow resetting, and adds specs.


steam:
Looks good. I think we'll want a decent amount of manual testing. Logging in, out, etc. 👍

PR #88, opened 2016-07-28
Fixes the cover image on iPad

colinta:
uses `original` instead of `hdpi`, and uses `xhdpi` instead of `hdpi` on iPhone, which should look much better on iPhone 6+

The `hdpi` cover image is actually pretty dang small.  This brings us closer in line w/ webapp (esp in landscape).

Also, this PR removes all references to `xxhdpi` and `xxxhdpi`, because the server doesn't generate those, and we'll never get them.

### DIFF
--- a/Resources/StubbedResponses/activity-own-post.json
+++ b/Resources/StubbedResponses/activity-own-post.json
@@ -69,24 +69,6 @@
                             "width": 1500,
                             "height": 641
                         }
-                    },
-                    "xxhdpi": {
-                        "url": "https://d1qqdyhbrvi5gr.cloudfront.net/uploads/asset/attachment/5381/xxhdpi.jpg",
-                        "metadata": {
-                            "size": 728689,
-                            "type": "image/jpeg",
-                            "width": 2560,
-                            "height": 1094
-                        }
-                    },
-                    "xxxhdpi": {
-                        "url": "https://d1qqdyhbrvi5gr.cloudfront.net/uploads/asset/attachment/5381/xxxhdpi.jpg",
-                        "metadata": {
-                            "size": 728687,
-                            "type": "image/jpeg",
-                            "width": 2560,
-                            "height": 1094
-                        }
                     }
                 }
             }
--- a/Resources/StubbedResponses/asset.json
+++ b/Resources/StubbedResponses/asset.json
@@ -11,6 +11,15 @@
                     "height": 1094
                 }
             },
+            "original": {
+                "url": "//example.com/5381/original.jpg",
+                "metadata": {
+                    "size": 728689,
+                    "type": "image/jpeg",
+                    "width": 2560,
+                    "height": 1094
+                }
+            },
             "small_screen": {
                 "url": "//example.com/5381/small_screen.jpg",
                 "metadata": {
@@ -54,24 +63,6 @@
                     "type": "image/jpeg",
                     "width": 1500,
                     "height": 641
-                }
-            },
-            "xxhdpi": {
-                "url": "https://example.com/5381/xxhdpi.jpg",
-                "metadata": {
-                    "size": 728689,
-                    "type": "image/jpeg",
-                    "width": 2560,
-                    "height": 1094
-                }
-            },
-            "xxxhdpi": {
-                "url": "https://example.com/5381/xxxhdpi.jpg",
-                "metadata": {
-                    "size": 728687,
-                    "type": "image/jpeg",
-                    "width": 2560,
-                    "height": 1094
                 }
             }
         }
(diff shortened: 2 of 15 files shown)


Resources/StubbedResponses/activity-own-post.json
```diff
@@ -69,24 +69,6 @@
                             "width": 1500,
                             "height": 641
                         }
-                    },
```
steam: Did this get removed as a result of re-running the rake task that generates these or was the code removed by hand?

Resources/StubbedResponses/asset.json
```diff
@@ -11,6 +11,15 @@
                     "height": 1094
                 }
             },
+            "original": {
```
steam: Same here, were they added by hand or auto-generated?

Sources/Model/User.swift
```diff
@@ -278,7 +278,12 @@ extension User {
         if animated && (!postsAdultContent || viewsAdultContent == true) && coverImage?.original?.url.absoluteString.endsWith(".gif") == true {
             return coverImage?.original?.url
         }
-        return coverImage?.hdpi?.url
+        if UIDevice.currentDevice().userInterfaceIdiom == .Pad {
+            return coverImage?.original?.url
```
steam: Is original resized at all? Are we concerned about these being way **too** big?

colinta: Good question for @jayzes or @rynbyjn

Resources/StubbedResponses/asset.json
colinta: All the JSON changes were by hand - I'll re-run the generator and see what we get.

steam: 👍 seems like a good plan. I'm not sure if the generators are maintained but it would be a bummer to have these re-added/removed unexpectedly in the future. @jayzes, @666, any thoughts?

rynbyjn: That generator should be working as expected still. I'm not 100% sure if it would remove these now, but they are definitely not in the asset uploaders in the [api](https://github.com/ello/ello/blob/master/app/uploaders/asset_uploader.rb#L33).

colinta:
@steam I did run the `rake generate:responses` task, and it generated TONS of differences, but `xxhdpi` did not reappear.

I don't want to commit that, though, since it ruins the specs in just about every OTHER way.

rynbyjn:
I remember talking about removing all of the tests around `id`, `createdAt`, and other dates. It seems to be most of what changes in the updated files when you run that command and doesn't really provide much value IMO.

Sources/Model/User.swift
steam: @jayzes @Justin-Holmes Any thoughts on this?

steam:
Gonna give this a thumbs up pending feedback from @jayzes or @Justin-Holmes on the image size question.

Sources/Model/User.swift
colinta: No word from these jokers, but looking at `Cover.js`, they have a simple rule: always use `xhdpi`.  I'll update this to use that rule.

colinta:
@mkitt showed me that they DO use other image sizes! <https://github.com/ello/webapp/blob/master/src/components/viewport/ResizeComponent.js#L25>

The iPad Pro has a width of 1024pt or 1366pt, which would register as `xhdpi` by this logic, though the `optimized` size is 1920 wide, which would be even better, but almost 3x the file size (~1.5mb)

My xhdpi cover image in xhdpi is ~500kb (<https://assets.ello.co/uploads/user/cover_image/548729/ello-xhdpi-99584e59.jpg>), which seems like a reasonable file size.  Its 1116px width is great for iphone 6+ (1080px or 1242px, depending on who you ask).

I'm leaning towards always using `xhdpi` (and this PR has been updated to do that).  We can always re-adjust later.

jayzes:
Sorry for the delay - that SGTM! I was going to defer to Matt on it anyway, haha.

--
Jay Zeschin

On August 2, 2016 at 2:08:42 PM, Colin T.A. Gray (notifications@github.com) wrote:

@mkitt showed me that they DO use other image sizes! https://github.com/ello/webapp/blob/master/src/components/viewport/ResizeComponent.js#L25

The iPad Pro has a width of 1024pt or 1366pt, which would register as xhdpi by this logic, though the optimized size is 1920 wide, which would be even better, but almost 3x the file size (~1.5mb)

My xhdpi cover image in xhdpi is ~500kb (https://assets.ello.co/uploads/user/cover_image/548729/ello-xhdpi-99584e59.jpg), which seems like a reasonable file size. Its 1116px width is great for iphone 6+ (1080px or 1242px, depending on who you ask).

I'm leaning towards always using xhdpi (and this PR has been updated to do that). We can always re-adjust later.

PR #90, opened 2016-08-01
Adds Buy Button to Omnibar

colinta:
- [x] add buy button appears in editor
- [x] add buy button icon to images
- [x] remove buy button link
- [x] edit post w/ buy button

Lots of "edge case" behavior, too, like: reordering disables the buy button, re-enables when reordering is complete.

Added lots of snapshot specs for the Omnibar.


Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
```diff
@@ -0,0 +1,191 @@
...
+    private func style() {
+        backgroundButton.backgroundColor = .modalBackground()
+
+        titleLabel.font = .defaultFont(18)
+        titleLabel.textColor = .whiteColor()
+
+        cancelLabel.font = .defaultFont()
+        cancelLabel.textColor = .greyA()
+
+        productLinkField.backgroundColor = .whiteColor()
```
steam: Maybe move the the `productLinkField` configuration into a helper function?

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        submitButton.snp_makeConstraints { make in
+            make.top.equalTo(productLinkField.snp_bottom).offset(10)
+            make.leading.equalTo(self).offset(10)
+            submitButtonTrailingRight = make.trailing.equalTo(self).offset(-10).constraint
+            submitButtonTrailingRemove = make.trailing.equalTo(removeButton.snp_leading).offset(-10).constraint
+            make.height.equalTo(60)
+        }
+        submitButtonTrailingRemove!.deactivate()
```
steam: Should the constraints be implicitly unwrapped when declared rather than force unwrapped when they're referenced?

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
```diff
@@ -0,0 +1,191 @@
...
+            make.top.equalTo(submitButton.snp_bottom).offset(25)
+            make.leading.equalTo(self).offset(10)
+        }
+    }
+
+    func productLinkDidChange() {
+        submitButton.enabled = isValidLink(productLinkField.text)
+    }
+
+    func isValidLink(urlString: String?) -> Bool {
```
steam: Would `isValidLink(:_)` be useful as an app wide utility?

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
```diff
@@ -0,0 +1,191 @@
...
+    func closeModal() {
+        delegate?.closeModal()
+    }
+
+    func submitLink() {
+        guard let urlString = productLinkField.text else {
+            return
+        }
+
+        var url: NSURL?
```
steam: The `NSURL` creation is similar to the way it is done in `isValidLink` above. Depending on what is done above these two functions could share a helper.

Sources/Controllers/Omnibar/AffiliateLinkViewController.swift
```diff
@@ -0,0 +1,50 @@
+////
+///  AffiliateLinkViewController.swift
+//
+
+public protocol AffiliateLinkControllerDelegate: class {
+    func submitAffiliateLink(url: NSURL)
+    func clearAffiliateLink()
+}
+
+public class AffiliateLinkViewController: UIViewController {
```
steam: Such a tidy little controller :metal:. The screen pattern really helps here. Nice work.

Sources/Controllers/Omnibar/OmnibarCacheData.swift
```diff
@@ -0,0 +1,25 @@
+////
+///  OmnibarCacheData.swift
+//
+
+public class OmnibarCacheData: NSObject, NSCoding {
```
steam: What motivated this class? Seems useful, just curious about your thought process.

Sources/Controllers/Omnibar/OmnibarImageCell.swift
```diff
@@ -28,8 +30,19 @@ public class OmnibarImageCell: UITableViewCell {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
 
         flImageView.clipsToBounds = true
-        flImageView.contentMode = .ScaleAspectFill
+        flImageView.contentMode = .ScaleAspectFit
         contentView.addSubview(flImageView)
+
+        affiliateButton.backgroundColor = .greenD1()
```
steam: Should the `affiliateButton` setup be up above instead of inside of `init`? i.e.
```
public let affiliateButton: UIButton = {
    let button = UIButton()
    // .. configure here
    return button
}()
```
I guess we've never settled one way or another on that. Maybe a helper function at the very least. Thoughts?

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
```diff
@@ -0,0 +1,191 @@
...
+
+import SnapKit
+
+public protocol AffiliateLinkDelegate: class {
+    func closeModal()
+    func submitLink(url: NSURL)
+    func clearLink()
+}
+
+public class AffiliateLinkScreen: UIView {
```
steam: As I was reviewing changes to `OmnibarImageCell` I was reminded of the `Size` struct at the top of the class. Should we adopt size structs for `XScreen` constraint dimensions? Something to consider.

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
colinta: OTOH, that creates more places to look for styling code.  This is a heftier chunk of styling code, but its function is the same as the other view-styling code in this method.

If the view code was in a controller, you'd have this problem of sprawling view code everywhere, and so separating it out helps readability.  In a Screen class, though, you're already focused on UI & interaction, so a larger "style all the views" method makes sense to me.

If we do want to move to `styleSomeView` methods, I'd like to take a look at all the Screen classes and change them en masse.  Otherwise this will be the one-off odd class that has a single, lonely style helper.

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
colinta: The `deactivate` method is supposed to be called (if I am reading the SnapKit code correctly) after `makeConstraints`, so I had to move this out of that block.

I'll refactor, I think a `let submitButtonTrailingRemove: Constraint` above that block will work.

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
colinta: Ooh good call, I meant to revisit these helpers.

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
steam: I like the idea of breaking out helpers and don't see a big conflict with other classes. Ideally we'd use them everywhere but I wouldn't want to not add them in a single place for fear of having to do it everywhere or nowhere. The smaller a function, the better imo. I'm not dogmatic about it though.

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
steam: 👍

Sources/Controllers/Omnibar/OmnibarViewController.swift
```diff
@@ -83,15 +83,15 @@ public class OmnibarViewController: BaseElloViewController, OmnibarScreenDelegat
...
-            return "omnibar_v2_comment_\(post.repostId ?? post.id)"
-        }
-        else if editPost != nil || editComment != nil {
-            return nil
-        }
-        else {
-            return "omnibar_v2_post"
+    override func didSetCurrentUser() {
+        super.didSetCurrentUser()
+        if isViewLoaded() {
```
steam: This seems like a good candidate for a `guard`.

Sources/Controllers/Omnibar/OmnibarViewController.swift
```diff
@@ -317,6 +330,40 @@ public class OmnibarViewController: BaseElloViewController, OmnibarScreenDelegat
...
+        }
+        self.presentViewController(controller, animated: true, completion: nil)
+    }
+
+    public func omnibarPushController(controller: UIViewController) {
+        self.navigationController?.pushViewController(controller, animated: true)
+    }
+
+    public func omnibarDismissController(controller: UIViewController) {
+        self.dismissViewControllerAnimated(true, completion: nil)
```
steam: Why does this function have the `controller` argument?

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
colinta: That refactor didn't work - I thought that the same mechanism that lets you assign a `let` constant in an `if` could work here, but it doesn't (just says "can't assign to a constant").

The force unwrap still feels right to me, because it's saying "this is assigned now", whereas optionally unwrapping says "I don't know".

Sources/Controllers/Omnibar/OmnibarCacheData.swift
colinta: It's been around since the early days.  It was called `OmnibarData`, and lived in `OmnibarViewController`.  It saves post and comment drafts.

Sources/Controllers/Omnibar/OmnibarImageCell.swift
colinta: I'm not a fan of that style, myself, I prefer the `style / setText / arrange` methods.  I'll refactor this.

steam: 👍

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
colinta: 👍 to this, will refactor.

Sources/Controllers/Omnibar/AffiliateLinkScreen.swift
steam: I was referring to declaring it as an implicitly unwrapped optional up top rather than force unwrapping it when it is used. Similar to the `IBOutlet` pattern.

colinta: Ahh, I didn't follow you there.  Yeah that would work here.

But *then* we have to discuss why I think `IBOutlet`s as declared by Apple are *all wrong* (they should be retained, not weak, optional on a controller, not force unwrapped 😝)

Sources/Controllers/Omnibar/OmnibarViewController.swift
colinta: No reason, afaik.  Removing.

colinta:
```Executed 2101 tests, with 0 failures (0 unexpected) in 126.943 (127.911) seconds```

steam:
Good to go. The github feature that allows you to view new changes is pretty awesome. Saved me a lot of time. :)

PR #91, opened 2016-08-02
Close the browser (and open externally) on failed URLs

colinta:
This covers itunes store links (and any other app urls)

### DIFF
--- a/Sources/Controllers/WebBrowser/ElloWebBrowserViewController.swift
+++ b/Sources/Controllers/WebBrowser/ElloWebBrowserViewController.swift
@@ -7,6 +7,7 @@ import Crashlytics
 
 public class ElloWebBrowserViewController: KINWebBrowserViewController {
     var toolbarHidden = false
+    var prevRequestURL: NSURL?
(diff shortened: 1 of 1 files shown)


Sources/Controllers/WebBrowser/ElloWebBrowserViewController.swift
```diff
@@ -76,7 +77,15 @@ public class ElloWebBrowserViewController: KINWebBrowserViewController {
 // MARK: ElloWebBrowserViewConteroller: KINWebBrowserDelegate
 extension ElloWebBrowserViewController: KINWebBrowserDelegate {
 
+    public func webBrowser(webBrowser: KINWebBrowserViewController!, didFailToLoadURL url: NSURL?, error: NSError!) {
+        if let url = url ?? prevRequestURL {
```
steam: I've never seen this syntax before. Does this assign `url` if it is present, otherwise it assigns `prevRequestURL` if is present? I dig it.

steam:
👍

PR #92, opened 2016-08-02
Fixes pagination, user loading, performance

colinta:
- [x] pagination could be triggered *immediately* after pulling to refresh - introduced `pagingEnabled` to prevent this
- [x] work on `UIWebView` that could only be performed on the main queue was being done on a worker queue
- [x] users were being loaded with 25 posts, now `loadUser` loads 0 posts, and `loadUserPosts` loads 10 posts


colinta:
Dammit, these failures are false negatives. `AsyncOperationSpec` has trouble with `cancelAllOperations()`, and I dunno why the post id is using the one that is loaded from the request (that request shouldn't be sent!).

steam:
Bummer on the specs. Looks good otherwise.

steam:
👍

PR #93, opened 2016-08-02
Add Segment tracking for buy button clicks

jayzes:
[#127593559]


jayzes:
@colinta @steam not sure if this is the right place to track this, but seemed reasonable?

colinta:
👍

PR #96, opened 2016-08-04
Fixes Following/Starred

colinta:
The new relationship was not being stored properly, so the "immediate UI update" wasn't working.


steam:
Come on Travis!

PR #97, opened 2016-08-04
Saw a crash on 'self.cellJobs.removeAtIndex(0)', so added a guard

colinta:


steam:
Seems legit. Good find.

PR #102, opened 2016-08-09
Adds loading states for Profile and Post Detail

colinta:
- [x] Shows placeholders during profile load (username, name, and post/following/followers/loves count)
- [x] Shows "image loading" animation on avatar and cover image
- [x] Shows "pull to refresh" spinner on post detail and under profile header while posts are loading
- [x] Lucian approved!™


colinta:
profileheadercell__while_loading__has_valid_snapshot_with_no_image 2x

colinta:
simulator screen shot aug 9 2016 2 22 03 pm
simulator screen shot aug 9 2016 2 22 23 pm

Sources/Controllers/Omnibar/OmnibarScreen.swift
```diff
@@ -426,9 +426,9 @@ public class OmnibarScreen: UIView, OmnibarScreenProtocol {
     }
 
     public func profileImageTapped() {
-        if let userParam = currentUser?.id {
-            let profileVC = ProfileViewController(userParam: userParam)
-            profileVC.currentUser = self.currentUser
+        if let currentUser = currentUser {
```
steam: This could become a `guard` instead of a regular `if let`.

Sources/Controllers/Profile/ProfileHeaderCell.swift
```diff
@@ -84,60 +126,69 @@ public class ProfileHeaderCell: UICollectionViewCell {
...
-            if user.id == currentUser?.id {
-                noResultsTitle = InterfaceString.Loves.CurrentUserNoResultsTitle
-                noResultsBody = InterfaceString.Loves.CurrentUserNoResultsBody
-            }
-            else {
-                noResultsTitle = InterfaceString.Loves.NoResultsTitle
-                noResultsBody = InterfaceString.Loves.NoResultsBody
-            }
-            simpleStreamDelegate?.showSimpleStream(.Loves(userId: user.id), title: InterfaceString.Loves.Title, noResultsMessages: (title: noResultsTitle, body: noResultsBody))
+        guard let user = user else { return }
```
steam: All these guards are so nice 👌

Sources/Controllers/Stream/Cells/StreamTextCell.swift
```diff
@@ -55,6 +55,7 @@ public class StreamTextCell: StreamRegionableCell, UIWebViewDelegate, UIGestureR
     override public func prepareForReuse() {
         super.prepareForReuse()
         webView.stopLoading()
+        webView.loadHTMLString("", baseURL: nil)
```
steam: For some reason I thought this caused a flicker, no?

steam:
A couple comments, otherwise. LGTM

Sources/Controllers/Omnibar/OmnibarScreen.swift
colinta: Noted 👍

Sources/Controllers/Stream/Cells/StreamTextCell.swift
colinta: We've def got issues around these UIWebViews.  I think a good solve would be to have a global "cache" of views, shared by the cells and the calculators, so that the loading of HTML only happens once per `TextRegion`, not N+1 times (1 for the calculator, and N for every time the cell comes into view).

I noticed, though, that the text was not updating - so a cell would have text from another post or comment, then would change to the new text.  It doesn't flicker any more, per se, it's just straight-up blank and the text "pops in", which felt better than having the old text there.

PR #103, opened 2016-08-10
Sets the buy button size according to list/grid setting

colinta:
40 for list, 30 for grid

Some trickiness in the constraints and views.  There is an invisible button, it's always 40x40 (to ensure a good hit area), and centered on the `buyButtonGreen` view.  It's the `buyButtonGreen` view that changes size.

The aspect ratio of `buyButtonGreen` is set to 1:1, so only the width constraint needs to be updated.


steam:
This seems like a good candidate for some snapshot specs. Worth it or are the snapshot specs a PITA right now?

colinta:
Snapshot specs are great, ever since we figured out the public / private pod issue.  Added snapshots.

colinta:
Unrelated spec failures, I'll look at em in another PR.

PR #104, opened 2016-08-12
Update pinned certs Cocoapod

jayzes:


Readme.md
```diff
@@ -44,72 +44,6 @@ Once you have staging and production client credentials, you can switch between
 - Provisioning Profile `iOSTeam Provisioning Profile: co.ello.ElloDev`
 
 
-### Universal Links
```
steam: Why was the universal link documentation removed?

jayzes: The only reason I moved that too is that it involves signing with the certs, which aren't in this project anymore. I could split this section though - leave the structure here and signing instructions there - if that's preferable?

steam: My only concern is "losing" it in the other repo. I'm not sure it'll be as easily accessible over there. My preference would be to leave the heart of the universal link instructions here and reference the signing instructions in the cert repo. :metal:

jayzes:
@steam updated in https://github.com/ello/ello-ios/pull/104/commits/610e6a0b4ad160872fe6413fa3da5a654d94c905

steam:
👍

PR #105, opened 2016-08-12
Adds Buy Button image to NotificationCell

colinta:
Just an image, not a button, so tapping still opens post detail.


steam:
:metal:

PR #106, opened 2016-08-12
Fix `NotificationCell` sizing and calculator code

colinta:
and only fire `onHeightMismatch` after everything (image & message HTMl) have loaded.  The `height` passed in is from `layoutSubviews`, so is guaranteed to be correct.

### DIFF
--- a/Sources/Controllers/Notifications/NotificationCell.swift
+++ b/Sources/Controllers/Notifications/NotificationCell.swift
@@ -17,23 +17,19 @@ public class NotificationCell: UICollectionViewCell, UIWebViewDelegate {
 
     struct Size {
         static let ButtonHeight = CGFloat(30)
-        static let ButtonMargin = CGFloat(15)
         static let WebHeightCorrection = CGFloat(15)
         static let SideMargins = CGFloat(15)
         static let AvatarSize = CGFloat(30)
         static let ImageWidth = CGFloat(87)
         static let InnerMargin = CGFloat(10)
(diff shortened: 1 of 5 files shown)


steam:
Looks like some specs are off now. Other than that 👍

Sources/Controllers/Notifications/NotificationCell.swift
```diff
@@ -17,23 +17,19 @@ public class NotificationCell: UICollectionViewCell, UIWebViewDelegate {
...
-        static let ButtonMargin = CGFloat(15)
         static let WebHeightCorrection = CGFloat(15)
         static let SideMargins = CGFloat(15)
         static let AvatarSize = CGFloat(30)
         static let ImageWidth = CGFloat(87)
         static let InnerMargin = CGFloat(10)
         static let CreatedAtMargin = CGFloat(-5)
-        static let CreatedAtHeight = CGFloat(10)
-
+        static let CreatedAtHeight = CGFloat(12)
```
steam: If you need to go back in to this code for the specs you could convert these to `static let NAME: CGFloat = X`. Not major but I've noticed you moving to that in other places.

colinta: Yup, actually I deliberately avoided doing that here, because it would cause more conflicts with #105 (it makes the same style changes).  I'll have to reconcile the changes to these values with the swift style changes either way.

PR #107, opened 2016-08-15
Support protocol relative image URLs by prepending 'https:'

colinta:


Sources/Controllers/Stream/Cells/StreamImageCell.swift
```diff
@@ -153,6 +153,12 @@ public class StreamImageCell: StreamRegionableCell {
     }
 
     private func loadImage(url: NSURL) {
+        guard url.scheme != "" else {
+            if let urlWithScheme = NSURL(string: "http:\(url.absoluteString)") {
```
steam: I wonder if `http` or `https` is a better default. Probably `http`, @jayzes what are your thoughts?

colinta: Actually, thinking about this again, a protocol-relative URL *implies* that the domain supports both `http` *and* `https`.  I'll switch it up unless @jayzes objects...?

jayzes: HTTPS SGTM

colinta:
Same two weird (unrelated) spec failures that we've seen before.

PR #110, opened 2016-08-18
Tweaks Profile reloading UI

colinta:
- make the interpolation a little more pronounced by changing the "from" color to C4C4C4 (was CCCCCC)
- reset the cover image (shows interpolating loading view) after reload


colinta:
@caseydoran I'd like to include these in 1.12

caseydoran:
@colinta - cool, we haven't done any QA yet so go for it it you want to get a new build out.

PR #115, opened 2016-08-19
Re-enables "tap text to open" in grid view

colinta:
Easy fix - the refactor to using `guard` resulted in a flipped conditional.


colinta:
The spec failure is the "post id" spec again, not a major concern (I think it's due to the post being pulled out of YAP)

PR #116, opened 2016-08-19
Notification cell calculator corrections

colinta:
Something about having a `<ul><li>` causes spacing to mess up - this essentially increases the spacing between the HTML message and the created at label.


colinta:
Legit spec failures (changes to calculator and notification cell layout).  Fixing.

colinta:
Fixing the size calculations (which are correct on my computer) led me down a deep, dark hole of memory leaks.  This branch is in the latest build, and so will prob be released soon.  Gonna merge and submit a follow up w/ spec and memory leak fixes.

PR #117, opened 2016-08-19
Fixes the "error" when showing a user with no posts

colinta:
The cast to `[Post]` was failing because "no content" is returned as `data: ""`.  This checks for `""` and the service replies with an empty posts array in that case.  The ProfileGenerator then calls `secondaryJSONAbleNotFound` on the destination, which shows the "This person hasn't posted yet" UI.

### DIFF
--- a/Sources/Controllers/Profile/ProfileViewController.swift
+++ b/Sources/Controllers/Profile/ProfileViewController.swift
@@ -234,6 +234,7 @@ public final class ProfileViewController: StreamableViewController {
     private func reloadEntireProfile() {
         coverImage.pin_cancelImageDownload()
         coverImage.image = nil
+        updateNoPostsView(show: false)
         generator?.load(reload: true)
     }
 
--- a/Sources/Controllers/Profile/ProfileViewController.swift
+++ b/Sources/Controllers/Profile/ProfileViewController.swift
@@ -87,7 +87,7 @@ public final class ProfileViewController: StreamableViewController {
         }
         postChangedNotification = NotificationObserver(notification: PostChangedNotification) { [unowned self] (post, change) in
             if post.authorId == self.currentUser?.id && change == .Create {
-                self.updateNoPostsView(false)
+                self.updateNoPostsView(show: false)
             }
         }
     }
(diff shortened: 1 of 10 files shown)


Sources/Controllers/Profile/ProfileViewController.swift
```diff
@@ -504,7 +503,6 @@ extension ProfileViewController:  StreamDestination {
 
     public func setPlaceholders(items: [StreamCellItem]) {
         streamViewController.clearForInitialLoad()
-        updateNoPostsView(items.count < 2)
```
colinta: Since "placeholders" are always the same number, I didn't understand why this was here at all - it will always be `false`, which hides the no posts view.  Instead, I moved this into "reload", and it's just an explicit `show: false`

Specs/Helpers/ElloSpecHelper.swift
```diff
@@ -29,6 +29,10 @@ class ElloConfiguration: QuickConfiguration {
             ElloProvider.shared.authState = .Authenticated
             ElloProvider.shared.queue = nil
             ElloProvider.sharedProvider = ElloProvider.StubbingProvider()
+
+            ElloLinkedStore.sharedInstance.writeConnection.readWriteWithBlock { transaction in
+                transaction.removeAllObjectsInAllCollections()
```
colinta: This reset fixes the "wrong post id" spec.

Specs/Model/PostSpec.swift
```diff
@@ -62,6 +62,8 @@ class PostSpec: QuickSpec {
 
                 let createdAtString = "2015-12-14T17:01:48.122Z"
                 let post = Post.fromJSON(parsedPost) as! Post
+                let author: User = stub(["id": post.authorId, "username": "archer"])
+                ElloLinkedStore.sharedInstance.setObject(author, forKey: post.authorId, inCollection: "users")
```
colinta: Because of `removeAllObjectsInAllCollections` the `post.author` was being destroyed between tests.

colinta:
Self reviewing.

PR #118, opened 2016-08-22
Fixes memory leaks and specs

colinta:


steam:
Nice work here Colin. I hope this keeps the app running longer/cleaner on folk's phones.

PR #119, opened 2016-08-23
Fixes the per-page param name

colinta:
- `/users/id` returns posts, and the number is determined by `post_count`
- `/users/id/posts` returns posts, and the number is determined by `per_page`

😐

### DIFF
--- a/Sources/Controllers/Profile/ProfileGenerator.swift
+++ b/Sources/Controllers/Profile/ProfileGenerator.swift
@@ -102,7 +102,6 @@ private extension ProfileGenerator {
     }
 
     func loadUserPosts(doneOperation: AsyncOperation) {
-        guard loadingToken.isValidInitialPageLoadingToken(localToken) else { return }
         let displayPostsOperation = AsyncOperation()
         displayPostsOperation.addDependency(doneOperation)
         queue.addOperation(displayPostsOperation)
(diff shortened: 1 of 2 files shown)


Sources/Controllers/Profile/ProfileGenerator.swift
```diff
@@ -102,7 +102,6 @@ private extension ProfileGenerator {
     }
 
     func loadUserPosts(doneOperation: AsyncOperation) {
-        guard loadingToken.isValidInitialPageLoadingToken(localToken) else { return }
```
steam: was this causing problems?

steam:
👍

Sources/Controllers/Profile/ProfileGenerator.swift
colinta: It could - it's preventing a request from *being sent*, but we want to prevent the *result*.  It's using `loadingToken` in a way that it wasn't intended for.

Plus, if you follow the bouncing ball, `loadingToken` is reset *immediately before* this method is called, in `load(reload:)`.  So by that logic this guard will *never* prevent anything.  But if it did prevent something, it would actually *cause* a bug.

So it's not doing anything by being there, and wouldn't help even if it did.

PR #123, opened 2016-08-25
adding a period to the discover popup text

caseydoran:


colinta:
I hope you didn't stay up all night working on this, @caseydoran.  Get some sleep!  I'll start reviewing these changes, but @steam I could use a hand.  @jayzes @Justin-Holmes any concerns from the API?

Looping in @rynbyjn and @mkitt, we might want to duplicate this behavior in the webapp.

PR #124, opened 2016-08-26
Adds new Startup, Join and Login screens

colinta:
Onboarding is next, but I wanted to gets :eyes: on these startup screens.


Sources/Controllers/Startup/StartupScreen.swift
```diff
@@ -0,0 +1,79 @@
...
+    @objc func signUpAction() {
+        delegate?.signUpAction()
+    }
+
+    @objc func loginAction() {
+        delegate?.loginAction()
+    }
+}
+
+extension StartupScreenProtocol {}
```
steam: What does this do?

Sources/Controllers/Stream/Cells/CategoryCell.swift
```diff
@@ -52,14 +52,12 @@ public class CategoryCell: UICollectionViewCell {
         contentView.addSubview(label)
 
         colorFillView.snp_makeConstraints { make in
-            make.top.equalTo(contentView)
-            make.left.equalTo(contentView)
+            make.top.leading.trailing.equalTo(contentView)
```
steam: This is where `SnapKit` really shines. Nice 👌

Sources/Controllers/Startup/StartupScreen.swift
colinta: Just follows the same pattern as the other controller/screen instances.  Defines the methods on the screen that the controller can call (in this case: none)

```
/Users/colinta/Code/ello/ello-ios/Sources/Controllers/Startup/StartupProtocols.swift:
    8  }
    9  
   10: protocol StartupScreenProtocol: class {
   11  }
   12  

/Users/colinta/Code/ello/ello-ios/Sources/Controllers/Startup/StartupScreen.swift:
   77  }
   78  
   79: extension StartupScreenProtocol {}
   80  

/Users/colinta/Code/ello/ello-ios/Sources/Controllers/Startup/StartupViewController.swift:
    4  
    5  public class StartupViewController: UIViewController {
    6:     var mockScreen: StartupScreenProtocol?
    7:     var screen: StartupScreenProtocol { return mockScreen ?? (self.view as! StartupScreenProtocol) }
    8      var parentAppController: AppViewController?
    9  
```

colinta: Oh!  I see - oops, that should be:

```swift
extension StartupScreen: StartupScreenProtocol {}
```

steam: 👍

Sources/Model/Tracker.swift
```diff
@@ -103,17 +103,17 @@ public extension Tracker {
         agent.track("tapped join from startup")
     }
 
-    func tappedSignInFromStartup() {
+    func tappedLoginFromStartup() {
```
steam: Should we change the tracking names to align with our new function names?

colinta: @jayzes?  I assumed *not to* because that would change our metrics (the "tapped sign in" event would drop to zero)

Sources/Views/ClearTextField.swift
```diff
@@ -0,0 +1,102 @@
+////
+///  ClearTextField.swift
+//
+
+public class ClearTextField: UITextField {
+    public let onePasswordButton = UIButton()
```
steam: Is it worth having a `OnePasswordButton` class or style defined in the new `ElloButton` pattern rather than configuring a generic `UIButton` in `sharedSetup()`?

colinta: Sure!  The `StyledButton` isn't in this PR, but I think a plain `UIButton` subclass that assigns the image and contentMode would be handy.

Sources/Views/ElloButton.swift
```diff
@@ -4,6 +4,7 @@
 
 import UIKit
 
+// black button, gray text, and sets the correct font
 public class ElloButton: UIButton {
```
steam: Not gonna spend time reviewing on `ElloButton` because I've 👀 the future. And it is 🔆.

colinta: Ha!  Yeah, my first stab at clearing this up was just these comments - but later realized that the whole system was just too top-heavy.

Sources/Views/Screens/Screen.swift
```diff
@@ -0,0 +1,84 @@
...
+        addSubview(keyboardAnchor)
+        backgroundColor = .whiteColor()
+
+        screenInit()
+        style()
+        bindActions()
+        setText()
+        arrange()
+
+        keyboardAnchor.snp_makeConstraints { make in
```
steam: Seems like this should go in `arrange()`. Left out so sub classes aren't required to call `super.arrange()`?

colinta: That was my thinking, yeah.

Sources/Views/Screens/StreamableScreen.swift
```diff
@@ -1,45 +1,26 @@
...
+    override func arrange() {
         addSubview(streamContainer)
         addSubview(navigationBar)
 
         navigationBar.snp_makeConstraints { make in
             let c = make.top.equalTo(self).constraint
             self.navigationBarTopConstraint = c.layoutConstraints.first!
-            make.left.equalTo(self)
-            make.right.equalTo(self)
+            make.leading.equalTo(self)
```
steam: `leading` and `trailing` could be chained if you're into that kind of thing... right?

Sources/Views/Screens/Screen.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    public required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func screenInit() {}
+    func style() {}
+    func bindActions() {}
+    func setText() {}
+    func arrange() {}
```
colinta: Since all these are empty, *strictly* speaking we don't care if `super` is called.  But I'm 100% ok with changing it to require super on all these, just for consistency.  In which case I'd move the keyboard layout code.

Sources/Views/Screens/Screen.swift
steam: 👍

Sources/Views/Screens/StreamableScreen.swift
colinta: Only if you prefer blue bike sheds 😛

Sources/Views/Screens/Screen.swift
steam: I suppose at this age in the code base thats a good idea. I really like how we're standardizing. Lets do it.

Sources/Views/Screens/StreamableScreen.swift
steam: I actually thought that you'd come back with something bike-shed-like. You had some refactors above that were chaining of `SnapKit` calls. I'm fine with both *_blue_* and **_ugly_** bike sheds, **_COLIN_**.

colinta: Usually I chain em, yeah.  But I don't particularly *care* if they're chained.  Plus I keep `git stash ; git checkout cg/new/startup ; git stash pop ; # make changes ...` for each change.

steam: 👌

steam:
Nice work on this, its a big'n. :metal:

rynbyjn:
Y'all should add Danger to this and it would tell you that it's a big PR :metal:

PR #125, opened 2016-09-08
Changes build settings to support "ElloDev" and "Ello" apps

colinta:
There is a new "ElloDev" scheme, which is configured to use the "Debug" setting on all builds - specifically Run and Archive.

Meanwhile, "Ello" scheme is the opposite - it uses the "Release" build config for Run and Archive.

The entitlements, bundle ID, and icon are all tied to the build config, so just using the correct Scheme will configure each app.

The only manual step is pointing to prod vs staging.


steam:
This XML is on point. 👈

PR #132, opened 2016-09-16
Changes to Onboarding

colinta:
- disables the 'Pick 3/2/1' and 'Invite People' buttons until they're green
- makes text fields easier to tap in Sign Up
- changes onboarding logic: show if version is `nil`, otherwise ignore.
- changes onboarding saving logic: updates version after setting the categories (even if you force quit the app)

### DIFF
--- a/Sources/Controllers/Onboarding/OnboardingScreen.swift
+++ b/Sources/Controllers/Onboarding/OnboardingScreen.swift
@@ -40,12 +40,11 @@ public class OnboardingScreen: EmptyScreen {
     override func style() {
         buttonContainer.backgroundColor = .greyE5()
         abortButton.hidden = true
-
         nextButton.hidden = true
     }
 
     override func bindActions() {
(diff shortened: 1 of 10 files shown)


Sources/Controllers/Onboarding/OnboardingScreen.swift
```diff
@@ -40,12 +40,11 @@ public class OnboardingScreen: EmptyScreen {
     override func style() {
         buttonContainer.backgroundColor = .greyE5()
         abortButton.hidden = true
-
         nextButton.hidden = true
     }
 
     override func bindActions() {
-        promptButton.addTarget(self, action: #selector(nextAction), forControlEvents: .TouchUpInside)
+        promptButton.enabled = false
```
steam: `promptButton.enabled = false` should probably be in `style()` or some other function other than `bindActions()`

steam: Per our IRL conversation this is good to go. 👍

PR #133, opened 2016-09-19
Adds notifyOfWatchesViaPush, notifyOfWatchesViaEmail user settings

colinta:
@Justin-Holmes Can you verify my JSON and local names are correct?  The JSON should match the server, and the local name should be the camel-cased version.

### DIFF
--- a/Sources/Model/User.swift
+++ b/Sources/Model/User.swift
@@ -5,7 +5,10 @@
 import Crashlytics
 import SwiftyJSON
 
-let UserVersion: Int = 2
+// version 1: initial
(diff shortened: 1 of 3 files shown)


Justin-Holmes:
LGTM!

Sources/Model/User.swift
```diff
@@ -5,7 +5,10 @@
 import Crashlytics
 import SwiftyJSON
 
-let UserVersion: Int = 2
+// version 1: initial
```
steam: Nice notes here @colinta 👌

PR #134, opened 2016-09-19
Fixes onboarding so user can finish last step

colinta:
The `promptButton` defaults to disabled, but I neglected to *re-enable* it during the last step.

@steam I could go either way on the placement of `enabled=false`.  Looking at it a second time, are you still leaning towards putting it in `style()`?


steam:
@colinta, style seemed odd to me as well, but I wasn't sure where interaction control should be. Seems fine for now.

PR #135, opened 2016-09-19
Validation fixes during onboarding

colinta:
Solves remaining QA issues:

- [x] if you type a name, then delete it, the button will become disabled again
- [x] if you type an invalid link and the server rejects it, you will be presented with the error message

Also I added an `OnboardingProceed` enum, which is passed into the `proceed()` closure to make it clear what should happen next (`.Continue`, `.Abort`, `.Error`).  These used to be `true / false / nil`, which are not very clear.


steam:
👍

PR #137, opened 2016-09-21
Watch post notifications

colinta:
- adds the watch-post notifications types to `Activity`
- displays them similarly as loves
- to share some code w/ loves, I added the `PostActionable` protocol, implemented by `Love` and `Watch`

### DIFF
--- a/Sources/Model/NotificationAttributedTitle.swift
+++ b/Sources/Model/NotificationAttributedTitle.swift
@@ -178,6 +178,48 @@ public struct NotificationAttributedTitle {
                 else {
                     return styleUser(author).append(styleText(" loved a repost of your post."))
                 }
+            case .WatchNotification:
+                if let love = subject as? Watch,
+                    post = love.post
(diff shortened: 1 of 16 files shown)


Sources/Model/NotificationAttributedTitle.swift
```diff
@@ -178,6 +178,48 @@ public struct NotificationAttributedTitle {
...
+                if let love = subject as? Watch,
+                    post = love.post
+                {
+                    return styleUser(author)
+                        .append(styleText(" watched your "))
+                        .append(stylePost("post", post))
+                        .append(styleText("."))
+                }
+                else {
+                    return styleUser(author).append(styleText(" watched your post."))
```
steam: I somehow missed that we were doing this. This seems like overkill from a feature point of view to me. The code looks fine. Just my thoughts about getting notifications that someone is watching a post of mine. Could be cool I guess.

PR #142, opened 2016-09-22
Implements new logging methods

colinta:
As part of trying to figure out the push notifications, I rewrote our `log` method.  Seems handy to have custom logs visible in debug builds, so I figured I'd put it in a PR!

Also: removes the logging of tokens, since we haven't had log out issues in a long time.

### DIFF
--- a/Sources/Networking/ProfileService.swift
+++ b/Sources/Networking/ProfileService.swift
@@ -125,7 +125,7 @@ public struct ProfileService {
     }
 
     public func updateUserDeviceToken(token: NSData) {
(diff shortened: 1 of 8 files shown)


Sources/Networking/ProfileService.swift
```diff
@@ -125,6 +125,7 @@ public struct ProfileService {
     }
 
     public func updateUserDeviceToken(token: NSData) {
+        log("push token", object: String(token.description.characters.filter { !"<> ".characters.contains($0) }))
```
steam: Woah, nice Elixir pattern match there @colinta

PR #145, opened 2016-09-22
Attempt to fix `StreamCreateCommentCell` stretching issues

colinta:
My theory is that changing the frame of `CreateCommentBackgroundView` wasn't resulting in a redraw, so I added `setNeedsDisplay`, and for good measure there is a `setNeedsLayout/layoutIfNeeded` combo in `init` as well.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -34,7 +34,7 @@ end
 
 def common_pods
   if ENV['ELLO_STAFF']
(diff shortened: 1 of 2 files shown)


steam:
This only has the ElloUIFonts 1.2 update in it for some reason.

colinta:
Ha, oops, I staged the commits and then *didn't* commit.

PR #150, opened 2016-09-26
Adds notifyOfCommentsOnPostWatchViaPush and notifyOfCommentsOnPostWatchViaEmail to User settings

colinta:
for inclusion in 1.14


Sources/Model/User.swift
```diff
@@ -8,6 +8,7 @@ import SwiftyJSON
 // version 1: initial
 // version 2: added isHireable
 // version 3: added notifyOfWatchesViaPush, notifyOfWatchesViaEmail
+// version 4: added notifyOfCommentsOnPostWatchViaPush, notifyOfCommentsOnPostWatchViaEmail
 let UserVersion: Int = 3
```
steam: Doesn't look like `UserVersion` was bumped to 4.

colinta:
Ah thx, yeah need to bump that int

PR #152, opened 2016-09-26
Fix Alert Presentation

steam:
Such a tiny thing, but it prevented our custom transitioning delegate from getting used. `presentingViewController` is now optional in iOS 10. We were forcing it to be present if the custom transitioning delegate were to be used. This PR fixes that and everything works again.

Also, local specs pass.

`Executed 2145 tests, with 0 failures (0 unexpected) in 218.625 (219.927) seconds`


Support/Info.plist
```diff
@@ -30,7 +30,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>4928</string>
+	<string>5076</string>
```
colinta: This tells me you built with `Ello` instead of `ElloDev`.  *I see everything!*

steam: Damn! I wondered how that go there. I archived this morning and was going to submit to apple but you were on it so I backed off. I was surprised to see that update but assumed it was correct. @colinta, the all seeing 👀

PR #154, opened 2016-09-27
Fixes to the `CategoryCardCell`

colinta:
So that 'selected' is only used on `.SelectableCategoryCard` types


Sources/Controllers/Stream/Cells/CategoryCardCell.swift
```diff
@@ -13,11 +13,21 @@ public class CategoryCardCell: UICollectionViewCell {
         static let selectedImageOffset: CGFloat = 5
     }
 
+    override public var selectable: Bool {
+        didSet { self.selected = selected }
```
colinta: @steam On rebase/recompile this caused an error for me!  Says "can't assign value to itself" or something like that.  I'll push a fix up.

PR #158, opened 2016-09-27
Fixes category card fixes

colinta:
Apparently this isn't allowed:

```swift
override public var selectable: Bool {
    didSet { self.selected = selected }
}
```

This code is better anyway, it's clearer, I think.


colinta:
```
Test Suite 'All tests' passed at 2016-09-27 14:58:16.699.
	 Executed 2203 tests, with 0 failures (0 unexpected) in 237.510 (238.550) seconds
```

PR #161, opened 2016-09-27
Adds 'invited friend during onboarding' tracking event

colinta:
We had a generic tracking event, need a specific one for onboarding

CC @Justin-Holmes


Justin-Holmes:
:metal:

PR #164, opened 2016-09-28
Don’t redraw rightBarButtonItems

steam:
iOS 10 is animating `rightBarButtonItems` when they're assigned during a pull-to-refresh on both Post Details and Profiles. Rather than reassigning the buttons on pull to refresh we'll show what's already there.

@colinta, this feels harmless enough to me, any negative side effects you can think of?

[fixes #131181209](https://www.pivotaltracker.com/story/show/131181209)

### DIFF
--- a/Sources/Controllers/Profile/ProfileViewController.swift
+++ b/Sources/Controllers/Profile/ProfileViewController.swift
@@ -24,6 +24,7 @@ public final class ProfileViewController: StreamableViewController {
     var deeplinkPath: String?
     var generator: ProfileGenerator?
     private var isSetup = false
+    private var rightButtonsInitialized = false
 
     @IBOutlet weak var navigationBar: ElloNavigationBar!
     @IBOutlet weak var whiteSolidView: UIView!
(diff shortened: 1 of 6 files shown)


colinta:
Looks good!  I had a thought last night, if you wanted to go back to array comparison, we could store the `UIBarButtonItem`s as instance variables, so that the built-in comparison WOULD work as expected.

Sources/Controllers/Profile/ProfileViewController.swift
```diff
@@ -292,9 +293,18 @@ public final class ProfileViewController: StreamableViewController {
...
+            elloNavigationItem.rightBarButtonItems = rightBarButtonItems
+            return
+        }
+
+        if !elloNavigationItem.areRightButtonsTheSame(rightBarButtonItems) {
+            elloNavigationItem.rightBarButtonItems = rightBarButtonItems
+        }
     }
 
+
```
colinta: There's an extra blank line here.

Sources/Controllers/Profile/ProfileViewController.swift
```diff
@@ -270,6 +270,7 @@ public final class ProfileViewController: StreamableViewController {
     }
 
     func assignRightButtons() {
+
```
colinta: I've noticed that sometimes you add a blank line after `func` definition and sometimes you don't, are you moving away from that style, or trying to keep it?

steam:
```
Test Suite 'All tests' passed at 2016-09-30 16:41:07.323.
	 Executed 2207 tests, with 0 failures (0 unexpected) in 243.682 (245.008) seconds
```

PR #165, opened 2016-09-28
Fixes the Watch notification text

colinta:
And adds the here-to-fore missing `watch_comment_notification` text.


Sources/Model/NotificationAttributedTitle.swift
```diff
@@ -179,37 +179,49 @@ public struct NotificationAttributedTitle {
...
+                        .append(stylePost("post", post))
+                        .append(styleText(" you’re watching."))
+                }
+                else {
+                    return styleUser(author).append(styleText(" commented on a post you’re watching."))
                 }
             case .WatchOnRepostNotification:
-                if let love = subject as? Watch,
-                    post = love.post
+                if let watch = subject as? Watch,
```
steam: You should ❤️ your ⌚ @colinta.

PR #166, opened 2016-09-29
WIP: Serialize collection view cell reloads

steam:
The generators, [ProfileGenerator](https://github.com/ello/ello-ios/blob/master/Sources/Controllers/Profile/ProfileGenerator.swift) and [PostDetailGenerator](https://github.com/ello/ello-ios/blob/master/Sources/Controllers/Stream/Detail/PostDetailGenerator.swift) are great... except when they are not. They speed up concurrent network request loading but have inadvertently affected other UI updates that occur in app. [StreamDataSource](https://github.com/ello/ello-ios/blob/master/Sources/Controllers/Stream/StreamDataSource.swift) is responsible for updating cells by reloading them when user interactions occur. We believe that loading of data, updating collection views and updates based on user interaction results in our data sources and collection views getting out of sync. 

This PR is the first step towards resynchronizing these three pieces of the system. The hope is that calls to reload index paths on the collection view will not double up by forcing them into an operation queue with a `maxConcurrentOperationCount` of 1. 

Thoughts @colinta?

### DIFF
--- a/Resources/Main.storyboard
+++ b/Resources/Main.storyboard
@@ -1,8 +1,8 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="10117" systemVersion="15G31" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES">
+<?xml version="1.0" encoding="UTF-8"?>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="11201" systemVersion="16A323" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
     <dependencies>
-        <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="10085"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="11161"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <scenes>
         <!--Ello Tab Bar Controller-->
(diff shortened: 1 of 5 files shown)


Resources/Main.storyboard
```diff
@@ -1,8 +1,8 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="10117" systemVersion="15G31" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES">
+<?xml version="1.0" encoding="UTF-8"?>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="11201" systemVersion="16A323" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
     <dependencies>
-        <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="10085"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="11161"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
```
colinta: Are all these storyboard changes intended?  Fine if they are, but they add noise if they're not.

Sources/Views/ElloCollectionView.swift
```diff
@@ -0,0 +1,27 @@
...
+            guard let sself = self else { return }
+            inForeground {
+                sself.performBatchUpdates({
+                    sself.reloadItemsAtIndexPaths(reloadPaths)
+                    }, completion: { _ in
+                        done()
+                })
+            }
+        })
+        queue.addOperation(operation)
```
colinta: Nice!  We could push this out, see if crashes decrease, then add to it w/ `removeItems` et al.

Resources/Main.storyboard
steam: Very noisy but there was one change needed (the reference to `ElloCollectionView`) so I decided to let all the bizarre crap that was changed be part of the PR.

Sources/Views/ElloCollectionView.swift
steam: Thats a good idea. Get it in a build today. I'll merge it if you don't have any other feedback or thoughts.

Resources/Main.storyboard
colinta: Ahh right.

PR #169, opened 2016-10-03
Adds `isCollaborateable` property to `User`

colinta:
Also added the button to `ProfileViewController`, but as an optional, until we add it to the upcoming `ProfileScreen`


Sources/Controllers/Profile/ProfileViewController.swift
```diff
@@ -311,6 +313,14 @@ public final class ProfileViewController: StreamableViewController {
         createPost(text: "\(user.atName) ", fromController: self)
     }
 
+    @IBAction func collaborateButtonTapped() {
+        guard let user = user else { return }
+
+        Tracker.sharedTracker.tappedHire(user)
```
steam: Should this be a new event? Something like `tappedCollaborate` maybe.

colinta:
OK, updated the tracking event - also added a `fatalError` as a form of "TODO" - the `HireViewController` needs to be updated to support collaboration.

colinta:
```Executed 2218 tests, with 0 failures (0 unexpected) in 294.558 (296.204) seconds```

PR #170, opened 2016-10-03
Adds `hasAutoWatchEnabled` property to `Profile`

colinta:
*AND* moves a bunch of "should be in Profile" properties out of `User`, into `Profile`.

This should be merged into `cg/add/is_collaborateable`, since it uses changes from that branch.


colinta:
```Executed 2218 tests, with 0 failures (0 unexpected) in 279.452 (281.398) seconds```

PR #177, opened 2016-10-07
Stubbed Profile Header Views

steam:
The intention here is to give us a foundation to work from. There are 6 `ProfileBaseView` subclasses as well as 6 size calculators. Currently the size calculators report back a hard coded value. We'll need to make them dynamic moving forward. Additionally, the presenter is _mostly_ commented out. This is in order get the app to compile. We'll need to add bits and pieces of the presenter back in as functionality is added to the `ProfileBaseViews`.

Ideally we'd merge this into `sd/feature/profiles-two-point-oh` before any other profiles work happens.


steam:
screen shot 2016-10-07 at 9 26 26 am

colinta:
My only comment/gripe is the names of some of these - before we etch them in stone can we toss around some alternatives for `ViewsView` and `ActivityView`? These two had me scratching my head afa what content they contained.

A little more self-documenting:

- TotalViewsView
- TotalCountView
- CountAndCategoryView

- ProfileStatsView
- ProfileCountsView

colinta:
And, since I'm being pedantic, how about `NameView` -> `NamesView`, since it contains username and full name.  (I called it `NamesView` in the branch where I started to work on that one)

Sources/Controllers/Profile/Calculators/ProfileLinksSizeCalculator.swift
```diff
@@ -0,0 +1,17 @@
+////
+///  ProfileLinksSizeCalculator.swift
+//
+
+import FutureKit
+
+
+public class ProfileLinksSizeCalculator: NSObject {
```
colinta: Why do they subclass `NSObject`?  Copy/paste artifact, or are we doing object identity or something?

Sources/Controllers/Profile/ProfileHeaderCell.swift
```diff
@@ -15,194 +15,213 @@ public protocol PostsTappedResponder {
...
+//    @IBOutlet weak var placeholders: UIView!
+//    @IBOutlet weak var nameLabel: ElloLabel!
+//    @IBOutlet weak var viewTopConstraint: NSLayoutConstraint!
+//    @IBOutlet weak var webViewHeightConstraint: NSLayoutConstraint!
+//    @IBOutlet weak var avatarWidthConstraint: NSLayoutConstraint!
+//    @IBOutlet weak var bioWebView: UIWebView!
+//    @IBOutlet weak var postsButton: TwoLineButton!
+//    @IBOutlet weak var followersButton: TwoLineButton!
+//    @IBOutlet weak var followingButton: TwoLineButton!
+//    @IBOutlet weak var lovesButton: TwoLineButton!
```
colinta: I vote that we `rm -f` this code.  If we need to reference it, we've got the git history.

Sources/Controllers/Stream/CellDequeing/ProfileHeaderCellPresenter.swift
```diff
@@ -17,85 +17,87 @@ public struct ProfileHeaderCellPresenter {
...
-                cell.lovesButton.enabled = false
-                cell.followersButton.enabled = false
-            }
         }
+
+//            cell.onWebContentReady { webView in
+//                let webViewHeight = webView.windowContentSize()?.height ?? 0
+//                let actualHeight = ProfileHeaderCellSizeCalculator.calculateHeightBasedOn(
+//                    webViewHeight: webViewHeight,
+//                    nameSize: cell.nameLabel.intrinsicContentSize(),
```
colinta: Same thing here - we've got github or `git checkout` to look at the old content, let's just dump it and green field it!

Sources/Controllers/Profile/Calculators/ProfileLinksSizeCalculator.swift
steam: Yes, an artifact due to the old `ProfileHeaderCellSizeCalculator` needing to be a `UIWebViewDelegate` I believe. I can remove them and make them structs for now.

Sources/Controllers/Profile/ProfileHeaderCell.swift
steam: 👍

Sources/Controllers/Stream/CellDequeing/ProfileHeaderCellPresenter.swift
steam: 👍, I'll do you one better and delete the cell storyboard as well!

colinta: Haha hell yeah!

steam:
The specs are definitely not going to pass on this one so I'm going to merge it so we can sail forth.

PR #179, opened 2016-10-10
Profile Avatar View

steam:
Layout and specs for the avatar section of `ProfileHeaderCell`. We'll likely want to take another pass at these once they're all done. 

screen shot 2016-10-10 at 3 25 40 pm


steam:
Executed 2239 tests, with 0 failures (0 unexpected) in 238.465 (239.840) seconds

colinta:
Fancy!  Reviewing now...

Sources/Controllers/Profile/Calculators/ProfileAvatarSizeCalculator.swift
```diff
@@ -9,7 +9,7 @@ public struct ProfileAvatarSizeCalculator {
 
     public func calculate(item: StreamCellItem) -> Future<CGFloat> {
         let promise = Promise<CGFloat>()
-        promise.completeWithSuccess(122)
+        promise.completeWithSuccess(255)
```
colinta: `255` should be written as `0b11111111` just because we can 😜

Sources/Controllers/Profile/ProfileHeaderCompactView.swift
```diff
@@ -6,7 +6,7 @@ public class ProfileHeaderCompactView: ProfileBaseView {
 
     public struct Size {
         static let avatarWidth: CGFloat = 122
-        static let avatarHeight: CGFloat = 122
+        static let avatarHeight: CGFloat = 255
```
colinta: `0xFF` (ok I'm done w/ that joke)

Sources/Controllers/Profile/Views/ProfileAvatarView.swift
```diff
@@ -4,48 +4,80 @@
...
 
         layoutIfNeeded()
     }
+
+    override public func layoutSubviews() {
+        super.layoutSubviews()
+        avatarImageView.layer.cornerRadius = Size.avatarWidth / 2
+    }
+
+    public func prepareForReuse() {
```
colinta: This could be called from `ProfileHeaderCompactView.swift`, or from `ProfileHeaderCell`

steam: Good call, I made a mental note to do this in the future but it would be good to get it going now. I'll add it to this PR.

Sources/Controllers/Profile/ProfileHeaderCompactView.swift
steam: :trollface:

Sources/Controllers/Profile/Views/ProfileAvatarView.swift
colinta: Cool!  I'll follow this pattern in the `ProfileStatsView` and update `NamesView`, too.

PR #180, opened 2016-10-10
Implements the `ProfileStatsView` and friends

colinta:
```
Test Suite 'All tests' passed at 2016-10-10 17:12:05.784.
   Executed 2249 tests, with 0 failures (0 unexpected) in 226.101 (227.078) seconds
```

ProfileStatsView


Sources/Controllers/Profile/Views/ProfileStatsView.swift
```diff
@@ -3,38 +3,151 @@
...
+    private let lovesCaptionLabel = UILabel()
+    private var captionLabels: [UILabel] {
+        return [postsCaptionLabel, followingCaptionLabel, followersCaptionLabel, lovesCaptionLabel]
+    }
+
+    private let postsButton = UIButton()
+    private let followingButton = UIButton()
+    private let followersButton = UIButton()
+    private let lovesButton = UIButton()
+    private var buttons: [UIButton] {
```
steam: I like this setup. Nice work. 👌
One thing we spoke about recently was providing visual feedback for UI interactions. Little things to help aid the perception of responsiveness. Decoupling the labels from the buttons makes that a bit harder. What do you think about adding `.TouchDown` and selected color changes to mimic WebApp's hover and selected states?

colinta: I dig it!

Sources/Controllers/Profile/Views/ProfileStatsView.swift
```diff
@@ -3,38 +3,176 @@
...
+        followingButton.addTarget(self, action: #selector(followingButtonTapped), forControlEvents: .TouchUpInside)
+        followersButton.addTarget(self, action: #selector(followersButtonTapped), forControlEvents: .TouchUpInside)
+        lovesButton.addTarget(self, action: #selector(lovesButtonTapped), forControlEvents: .TouchUpInside)
+
+        postsButton.addTarget(self, action: #selector(buttonDown(_:)), forControlEvents: [.TouchDown, .TouchDragEnter])
+        followingButton.addTarget(self, action: #selector(buttonDown(_:)), forControlEvents: [.TouchDown, .TouchDragEnter])
+        followersButton.addTarget(self, action: #selector(buttonDown(_:)), forControlEvents: [.TouchDown, .TouchDragEnter])
+        lovesButton.addTarget(self, action: #selector(buttonDown(_:)), forControlEvents: [.TouchDown, .TouchDragEnter])
 
+        postsButton.addTarget(self, action: #selector(buttonUp(_:)), forControlEvents: [.TouchUpInside, .TouchCancel, .TouchDragExit])
```
steam: Nice approach. For some reason I've never thought to have two separate `.TouchUpInside` selectors. Keeps the logic simple and isolated.

colinta: And reusable in this case - the button gets passed in, and I adjust the labels that "match" it.  But the no.1 reason is to support the `.Cancel | .DragExit` events - just `.UpInside` would mean that you could drag your finger out and it would stay highlighted, even after you lift your finger.

steam: Yep, makes sense. My initial approach would have been more complicated. I'd have 1 selector for `.Cancel | .DragExit`, similar to how you have it and then I'd have repeated code in the existing `.UpInside` selector. No bueno. Your solution is much more elegant. 👌

colinta:
good to merge?

steam:
Yep, I marked it "approved" in GitHub land but forgot to change the label. GTG.

PR #183, opened 2016-10-11
Implements `ProfileBioView`

colinta:


steam:
Bummer, more of the async spec failures.

Sources/Controllers/Profile/Calculators/ProfileBioSizeCalculator.swift
```diff
@@ -5,13 +5,46 @@
 import FutureKit
 
 
-public struct ProfileBioSizeCalculator {
+public class ProfileBioSizeCalculator: NSObject {
+    let webView = UIWebView()
+    let promise = Promise<CGFloat>()
+    var strongSelf: ProfileBioSizeCalculator?
```
steam: Do we do this anywhere else in the app (at the class/type level)? I'm wondering if this is best kept alive at a higher level by `ProfileHeaderCellSizeCalculator`.

colinta: Removed!

PR #184, opened 2016-10-12
Profile Total View Count

steam:
Adds the UI for Total View Count. User interaction and category modal will be a separate PR.

profiletotalcountview_on_custom_sized__375_0__60_0_ 2x


steam:
Just realized that the category badge will need to be hidden when the user is not featured in any categories. I'll include that in a new PR that includes the featured category modal.

Sources/Model/User.swift
```diff
@@ -10,6 +10,7 @@ import SwiftyJSON
 // version 3: added notifyOfWatchesViaPush, notifyOfWatchesViaEmail
 // version 4: added notifyOfCommentsOnPostWatchViaPush, notifyOfCommentsOnPostWatchViaEmail
 // version 5: added isCollaborateable, moved notifyOf* into Profile
+// version 6: added totalViewCount, categories
```
colinta: Mentions `categories`, but they're not added here.

steam: I'll kill it for now.

steam: Killed it.

PR #185, opened 2016-10-14
Adds the "Collab" button and updates `RelationshipControl`

colinta:
Couldn't have one without the other, because adding the Collab button requires the tighter layout of the new follow/star design.

`FollowButton` and `StarButton` are now `StyledButton` subclasses (a new departure: previously, `StyledButton` was a `final` class), and they update their style and images when the relationship changes.  I think it's easier to follow how the `relationshipControlStyle` (`.Default` or `.ProfileView`) and `relationshipPriority` affect the button colors/icon.

The `HireViewController` wasn't renamed, but I added the `UserEmailType` (`.Collaborate` or `.Hire`) to control the title and endpoint. We could rename the controller to match, but it didn't feel necessary at the time.

Some refactoring of the `ProfileScreen` to make it more similar to the "modern" screens like `LoginScreen`, but also modified the constraints to work with new margins and the showing/hiding of all the user buttons (collab, hire, and mention).


Sources/Controllers/Hire/HireViewController.swift
```diff
@@ -2,18 +2,33 @@
...
-    required public init(user: User) {
+    required public init(user: User, type: UserEmailType) {
         self.user = user
+        self.contactType = type
         super.init(nibName: nil, bundle: nil)
 
-        title = "Email \(user.atName)"
+        switch contactType {
+        case .Hire:
+            title = "Hire \(user.atName)"
```
steam: Should these be in `InterfaceString`?

colinta:
Good call - made that l18n change.

PR #187, opened 2016-10-17
Implements the `ProfileHeaderCompactView` layout

colinta:
Introduces `CalculatedCellHeights` struct to store all the calculated sizes.  Lots of red/green noise around this change.

Some refactors to constraints in `ProfileHeaderCompactView` (added `centerX` constraints, otherwise just moved some things around).

Fixed `ProfileBaseView` so that it doesn't call `privateInit` twice.

### DIFF
--- a/Sources/Controllers/Profile/ProfileHeaderCompactView.swift
+++ b/Sources/Controllers/Profile/ProfileHeaderCompactView.swift
@@ -22,12 +22,12 @@ public class ProfileHeaderCompactView: ProfileBaseView {
 
     var calculatedCellHeights: CalculatedCellHeights? {
         didSet {
-            if let calculatedCellHeights = calculatedCellHeights {
-                if let namesHeight = calculatedCellHeights.profileNames { namesHeightConstraint.updateOffset(namesHeight) }
-                if let bioHeight = calculatedCellHeights.profileBio { bioHeightConstraint.updateOffset(bioHeight) }
-                if let linksHeight = calculatedCellHeights.profileLinks { linksHeightConstraint.updateOffset(linksHeight) }
-                setNeedsLayout()
-            }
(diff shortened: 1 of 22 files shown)


Sources/Controllers/Profile/ProfileHeaderCompactView.swift
```diff
@@ -51,41 +55,40 @@ extension ProfileHeaderCompactView {
...
         }
 
         bioView.snp_makeConstraints { make in
-            make.width.equalTo(self.snp_width)
-            make.height.equalTo(Size.bioHeight)
             make.top.equalTo(self.statsView.snp_bottom)
+            make.width.centerX.equalTo(self)
+            bioHeightConstraint = make.height.equalTo(102).constraint
         }
+        bioHeightConstraint.updateOffset(103)
```
colinta: Oops, this was for debugging - it has no effect, but I'll remove it in another PR.

Sources/Controllers/Profile/ProfileHeaderCompactView.swift
```diff
@@ -51,41 +55,40 @@ extension ProfileHeaderCompactView {
...
+            make.top.width.centerX.equalTo(self)
             make.height.equalTo(Size.avatarHeight)
         }
 
         namesView.snp_makeConstraints { make in
-            make.width.equalTo(self.snp_width)
-            make.height.equalTo(Size.nameHeight)
             make.top.equalTo(self.avatarView.snp_bottom)
+            make.width.centerX.equalTo(self)
+            namesHeightConstraint = make.height.equalTo(101).constraint
```
colinta: Magic number was for debugging only - could/should be `0` or just not set at all.

colinta:
Dunno why I even promised another PR - I pushed that change here!  Should be GTG now.

PR #192, opened 2016-10-19
Empty bio / links

colinta:
Hide the bio or links if they're empty, and hide the gray lines appropriately.


Sources/Controllers/Profile/Views/ProfileBioView.swift
```diff
@@ -17,6 +17,10 @@ public class ProfileBioView: ProfileBaseView {
     }
     private let bioView = UIWebView()
     private let grayLine = UIView()
+    var grayLineVisible: Bool {
```
steam: Setting the cell's height to 0 doesn't hide this?

colinta: Different gray line - if the links are hidden, this line needs to be *hidden*, so that there isn't a "trailing line".

PR #193, opened 2016-10-19
Uses a user’s actual categories

steam:
The featured Category badge displays when a user is included in one or more categories.

### DIFF
--- a/Sources/Controllers/Stream/CellDequeing/ProfileTotalCountPresenter.swift
+++ b/Sources/Controllers/Stream/CellDequeing/ProfileTotalCountPresenter.swift
@@ -13,5 +13,6 @@ public struct ProfileTotalCountPresenter {
         currentUser: User?)
     {
         view.count = user.formattedTotalCount()
+        view.badgeButton.hidden = !(user.categories?.count > 0)
     }
 }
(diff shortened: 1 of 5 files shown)


Sources/Controllers/Stream/CellDequeing/ProfileTotalCountPresenter.swift
```diff
@@ -13,5 +13,6 @@ public struct ProfileTotalCountPresenter {
         currentUser: User?)
     {
         view.count = user.formattedTotalCount()
+        view.badgeButton.hidden = !(user.categories?.count > 0)
```
colinta: I think a getter/setter that doesn't expose the button would be appropriate here.

Specs/Controllers/Stream/CellDequeing/ProfileTotalCountPresenterSpec.swift
```diff
@@ -11,19 +11,39 @@ class ProfileTotalCountPresenterSpec: QuickSpec {
...
             }
+
+            it("shows cateogry badge if user is featured in one or more catgegories") {
+                let category: Ello.Category = Ello.Category.stub(["id" : "1", "name" : "art"])
+                let categories = [ category ]
+                let user = User.stub(["categories" : categories])
+                let view = ProfileTotalCountView()
+                ProfileTotalCountPresenter.configure(view, user: user, currentUser: nil)
+
+                expect(view.badgeButton.hidden) == false
```
colinta: and `view.badgeButtonVisible` here, of course.

PR #196, opened 2016-10-20
Fixes to `StreamTextCell`

colinta:
Sizing w/ multiple columns, and overlapping image/text.  Though, TBH, not entirely sure *why* that is fixed now.

### DIFF
--- a/Sources/Extensions/UIWebView.swift
+++ b/Sources/Extensions/UIWebView.swift
@@ -8,7 +8,7 @@ public extension UIWebView {
         if let jsWidth = self.stringByEvaluatingJavaScriptFromString("(document.getElementById('post-container') || document.getElementByTagName('div')[0] || document.body).offsetWidth")
             where !jsWidth.isEmpty
         {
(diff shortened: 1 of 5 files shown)


Sources/Extensions/UIWebView.swift
```diff
@@ -8,7 +8,7 @@ public extension UIWebView {
         if let jsWidth = self.stringByEvaluatingJavaScriptFromString("(document.getElementById('post-container') || document.getElementByTagName('div')[0] || document.body).offsetWidth")
             where !jsWidth.isEmpty
         {
-            if let jsHeight = self.stringByEvaluatingJavaScriptFromString("window.contentHeight()")
+            if let jsHeight = self.stringByEvaluatingJavaScriptFromString("(document.getElementById('post-container') || document.getElementByTagName('div')[0] || document.body).offsetHeight + 15")
```
steam: Why did this change?

colinta: The function was not defined in all cases!  I dug into one of the times where the `height` returned 0, and it turned out it was this JS code.  `window.contentHeight` was `undefined` (but only *sometimes*).

PR #197, opened 2016-10-20
Wire up bio and links views to open external links

steam:
Unfortunately we weren't easily able to use the responder chain in this case. `WebLinkDelegate` uses non Objective-C parameters in it's single function (`ElloAPI`). That prevents us from using a selector which must be exposed to Objective-C. Instead we're using the old `webLinkDelegate` pattern of assignment in `StreamDataSource` via a passthrough into both `ProfileBioView` and `ProfileLinksView`.

### DIFF
--- a/Sources/Controllers/Profile/ProfileHeaderCell.swift
+++ b/Sources/Controllers/Profile/ProfileHeaderCell.swift
@@ -38,7 +38,14 @@ public class ProfileHeaderCell: UICollectionViewCell {
     var bioView: ProfileBioView { get { return headerView.bioView } }
     var linksView: ProfileLinksView { get { return headerView.linksView } }
 
-    weak var webLinkDelegate: WebLinkDelegate?
+    weak var webLinkDelegate: WebLinkDelegate? {
+        set {
+            bioView.webLinkDelegate = newValue
+            linksView.webLinkDelegate = newValue
+        }
+        get { return bioView.webLinkDelegate }
+    }
+
     weak var simpleStreamDelegate: SimpleStreamDelegate?
     weak var userDelegate: UserDelegate?
 
--- a/Sources/Controllers/Profile/Views/ProfileLinksView.swift
+++ b/Sources/Controllers/Profile/Views/ProfileLinksView.swift
@@ -21,10 +21,13 @@ public class ProfileLinksView: ProfileBaseView {
         }
     }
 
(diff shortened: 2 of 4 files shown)


Sources/Controllers/Profile/ProfileHeaderCell.swift
```diff
@@ -38,7 +38,14 @@ public class ProfileHeaderCell: UICollectionViewCell {
     var bioView: ProfileBioView { get { return headerView.bioView } }
     var linksView: ProfileLinksView { get { return headerView.linksView } }
 
-    weak var webLinkDelegate: WebLinkDelegate?
+    weak var webLinkDelegate: WebLinkDelegate? {
+        set {
+            bioView.webLinkDelegate = newValue
```
steam: Assigns the webLinkDelegate to both bio and links.

Sources/Controllers/Profile/Views/ProfileLinksView.swift
```diff
@@ -135,8 +142,20 @@ extension ProfileLinksView {
         return button
     }
 
+    func buttonTapped(button: UIButton) {
+        guard let
+            externalLink = buttonLinks[button]
```
steam: hooray for fast lookup dictionaries

Sources/Controllers/Profile/Views/ProfileLinksView.swift
```diff
@@ -135,8 +142,20 @@ extension ProfileLinksView {
...
+        else { return }
+
+        let request = NSURLRequest(URL: externalLink.url)
+        ElloWebViewHelper.handleRequest(request, webLinkDelegate: webLinkDelegate)
```
steam: Use existing functionality defined in the web view helper.

Sources/Controllers/Stream/StreamDataSource.swift
```diff
@@ -327,9 +327,9 @@ public class StreamDataSource: NSObject, UICollectionViewDataSource {
             (cell as! NotificationCell).webLinkDelegate = webLinkDelegate
             (cell as! NotificationCell).userDelegate = userDelegate
             (cell as! NotificationCell).delegate = notificationDelegate
-//        case .ProfileHeader:
+        case .ProfileHeader:
```
steam: So sad, we need to use the delegate pattern here due to the above mentioned Swift/Objective-C interoperation issue.

PR #199, opened 2016-10-24
Remove Resources and About from Drawer

caseydoran:


caseydoran:
@steam not sure why this failed, any ideas (no rush)?

PR #201, opened 2016-10-31
1.17 fixes

steam:
Visual tweaks for 1.17


ShareExtension/ShareExtension.entitlements
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>aps-environment</key>
+	<string>development</string>
```
colinta: Should this be `production`, since the same entitlements file is used on `Ello` and `ElloDev` schemes?  Xcode, for some reason, always prompts to use `development`.

PR #202, opened 2016-10-31
Implements fixes for iPad, spec fixes

colinta:
For iPad (and larger phones), we support a larger `ProfileAvatarView` (it scales proportional to the `ratio` and `frame.width`)

Orientation changes will trigger a height mismatch in `ProfileAvatarView`.


colinta:
```
Test Suite 'All tests' failed at 2016-10-31 15:17:37.026.
	 Executed 2327 tests, with 27 failures (0 unexpected) in 252.240 (253.334) seconds
```

The failures are from `ProfileScreen` and `ProfileTotalCountView` snapshots.

PR #205, opened 2016-11-04
Implements `PagePromotional`s

colinta:
If the `Category` handed to `CategoryViewController` is a `level == .Meta` category, the `page_promotionals` endpoint will be fetched, and a random promotional will be selected as the "primary jsonable" for the category.

New cell type `PagePromotionalHeader`, and renamed `CategoryHeader` to `CategoryPromotionalHeader`.

### DIFF
--- a/Sources/Controllers/Categories/CategoryGenerator.swift
+++ b/Sources/Controllers/Categories/CategoryGenerator.swift
@@ -8,7 +8,11 @@ public final class CategoryGenerator: StreamGenerator {
     public var streamKind: StreamKind
     weak public var destination: StreamDestination?
 
-    private var category: Category?
+    private var category: Category
+    private var categoryIsMeta: Bool {
+        return category.level == .Meta
+    }
+    private var pagePromotional: PagePromotional?
     private var posts: [Post]?
     private var hasPosts: Bool?
     private var localToken: String!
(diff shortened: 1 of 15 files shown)


colinta:
@steam This could wait until specs are ready, and be rebased then.  At that point I'd like to rename `CategoryHeaderCellPresenter` to `CategoryPromotionalHeaderCellPresenter`.

Sources/Controllers/Categories/CategoryGenerator.swift
```diff
@@ -8,7 +8,11 @@ public final class CategoryGenerator: StreamGenerator {
     public var streamKind: StreamKind
     weak public var destination: StreamDestination?
 
-    private var category: Category?
+    private var category: Category
+    private var categoryIsMeta: Bool {
```
steam: We could move this to `Category.swift` and use it here as well as in `CategoryViewController.swift`.

Sources/Controllers/Categories/CategoryViewController.swift
```diff
@@ -133,17 +137,21 @@ extension CategoryViewController: StreamDestination {
...
+            else {
+                categoryPromotional = category.randomPromotional
+            }
+
+            self.title = category.name
         }
-        else {
-            categoryPromotional = category.randomPromotional
+        else if let pagePromotional = jsonable as? PagePromotional {
+            self.pagePromotional = pagePromotional
```
steam: Do we have the same potential issue of pull-to-refresh image swapping with page promotionals that we fixed for categoryPromotionals above?

colinta: Actually, the promotional is stored in the generator (along with the category).  I added it here for consistency only (it's not used anywhere in the controller).  On reload a new promotional will be displayed.

Sources/Controllers/Categories/CategoryGenerator.swift
colinta: I like it!

Sources/Controllers/Categories/CategoryViewController.swift
steam: 👍

steam:
@colinta - good to merge?

PR #207, opened 2016-11-09
Adds categories card list view to CategoryViewController

colinta:
In order to get the "fetch all categories" request into the `CategoryViewController`, I added a protocol `CategoryStreamDestination` which extends `StreamDestination`.  It's a private var on `CategoryGenerator`, and the `destination` returns that object, cast as a `StreamDestination`.  The swift compiler isn't quite smart enough to see that `CategoryStreamDestination` *satisfies* the `StreamDestination` type requirement, so this little handshake is required there.  Not a big fan, I wish we could just use `var destination: CategoryStreamDestination?`, but oh well.

Once the categories are loaded, we check for "did the user scroll?", and if not it adjusts the content inset so that the grid/list toggle is hidden.  The toggle doesn't start out hidden, but that enhancement probably wouldn't be terribly hard to add (in the "primary jsonable loaded" code?).

No specs yet, need @steam's work on specs to add specs to this feature.

### DIFF
--- a/Sources/Controllers/Categories/CategoryViewController.swift
+++ b/Sources/Controllers/Categories/CategoryViewController.swift
@@ -14,6 +14,7 @@ public final class CategoryViewController: StreamableViewController {
     var pagePromotional: PagePromotional?
     var categoryPromotional: Promotional?
     var generator: CategoryGenerator?
+    var userDidScroll: Bool = false
 
     public init(category: Category) {
         self.category = category
@@ -36,6 +37,7 @@ public final class CategoryViewController: StreamableViewController {
 
         let screen = CategoryScreen()
         screen.navigationItem = elloNavigationItem
+
         self.view = screen
         viewContainer = screen.streamContainer
         screen.delegate = self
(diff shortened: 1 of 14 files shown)


Sources/Controllers/Categories/CategoryViewController.swift
```diff
@@ -154,14 +163,41 @@ extension CategoryViewController: StreamDestination {
...
+            return CategoryCardListView.CategoryInfo(title: category.name, imageURL: category.tileURL, endpoint: .Category(slug: category.slug), selected: false)
+        }
+        screen.setCategoriesInfo(info, animated: true)
+        if let selectedCategoryIndex = selectedCategoryIndex {
+            screen.scrollToCategoryIndex(selectedCategoryIndex)
+        }
+        updateInsets()
+
+        if !userDidScroll && streamViewController.dataSource.visibleCellItems.count > 0 {
+            // streamViewController.collectionView.scrollToItemAtIndexPath(NSIndexPath(forItem: 1, inSection: 0), atScrollPosition: [.Bottom], animated: true)
```
steam: Can this be removed?

Sources/Networking/CategoryService.swift
```diff
@@ -9,15 +9,20 @@ public typealias CategoriesCompletion = (categories: [Category]) -> Void
 
 public class CategoryService {
 
-    public func loadCategories(success: CategoriesCompletion, failure: ElloFailureCompletion = { _ in }) {
+    public func loadCategories() -> Future<[Category]> {
```
steam: 👍 Nice **Future**

Sources/Controllers/Categories/CategoryViewController.swift
colinta: Caught me at my own "remove the comments!" game!

colinta:
Removed comment, and merging.

PR #208, opened 2016-11-09
Category Header Specs

steam:


Specs/Helpers/Stubs.swift
```diff
@@ -505,5 +558,36 @@ extension Ello.Category: Stubbable {
...
+        if let promotionals = values["promotionals"] as? [Promotional] {
+            var promotionalIds = [String]()
+            for promotional in promotionals {
+                promotionalIds.append(promotional.id)
+                ElloLinkedStore.sharedInstance.setObject(promotional, forKey: promotional.id, type: .PromotionalsType)
+            }
+            category.addLinkArray("promotionals", array: promotionalIds, type: .PromotionalsType)
+        }
+
+        if let body = values["body"] as? String {
```
colinta: These properties are all optionals anyway, so this could be made terser (and seems like we do this above/below):

```
category.body = values["body"] as? String
category.header = values["header"] as? String
```

steam: Good catch, I'll change them 👍

colinta:
👍 aww yeah

PR #214, opened 2016-11-11
Make See All root of Discover

steam:
This pr primarily replaces a post based discover landing screen with the category tile view. Additionally, it enables slug based loading of categories and "meta" categories. There is a still a bit of work to do to handle deep linking to a category in `AppViewController.swift` and `StreamViewController.swift`.


Sources/Controllers/App/AppViewController.swift
```diff
@@ -452,19 +452,20 @@ extension AppViewController {
...
-                nav.popToRootViewControllerAnimated(false)
-                discoverViewController.showCategory(data)
-            }
+            showCategoryScreen(data)
+//            if let nav = vc.selectedViewController as? UINavigationController,
+//                discoverViewController = nav.childViewControllers[0] as? DiscoverViewController
+//            {
+//                nav.popToRootViewControllerAnimated(false)
+//                discoverViewController.showCategory(data)
+//            }
```
colinta: Looks like the new code is implemented - unnecessary comments?  Remove in next PR?

steam: Yeah, this will get cleaned up in the next PR.

colinta: 👍 I'll just do a quick sanity check in this PR, then.

Sources/Utilities/Preloader.swift
```diff
@@ -11,8 +11,9 @@ public struct Preloader {
 
     public init(){}
 
-    public func preloadImages(jsonables: [JSONAble]) {
-
+    public func preloadImages(jsonables: [JSONAble]?) {
+        guard let jsonables = jsonables else { return }
```
steam: Good catch. I started to go down a different path and changed my mind. I'll revert this.

PR #216, opened 2016-11-14
Support new category deep linking

steam:
Deep linking to the new Category screens is setup. We may need to change this behavior pending feedback from Product but it at least functional.

The large number of lines is deceiving. Most of them come from `category.json`. Specs should pass this go around.

### DIFF
--- a/Sources/Controllers/Categories/CategoryViewController.swift
+++ b/Sources/Controllers/Categories/CategoryViewController.swift
@@ -18,9 +18,10 @@ public final class CategoryViewController: StreamableViewController {
     var generator: CategoryGenerator?
     var userDidScroll: Bool = false
 
-    public init(slug: String) {
+    public init(slug: String, name: String? = nil) {
         self.slug = slug
         super.init(nibName: nil, bundle: nil)
+        self.title = name
     }
 
     required public init?(coder aDecoder: NSCoder) {
@@ -28,7 +29,7 @@ public final class CategoryViewController: StreamableViewController {
     }
 
     override public func loadView() {
-        self.title = category?.name ?? slug.uppercaseFirst
+        self.title = category?.name ?? DiscoverType.fromURL(slug)?.name
         elloNavigationItem.title = title
         let item = UIBarButtonItem.backChevronWithTarget(self, action: #selector(backTapped(_:)))
         elloNavigationItem.leftBarButtonItems = [item]
--- a/Sources/Controllers/Categories/CategoryViewController.swift
+++ b/Sources/Controllers/Categories/CategoryViewController.swift
@@ -28,7 +28,7 @@ public final class CategoryViewController: StreamableViewController {
     }
 
     override public func loadView() {
-        self.title = category?.name
+        self.title = category?.name ?? slug.uppercaseFirst
         elloNavigationItem.title = title
         let item = UIBarButtonItem.backChevronWithTarget(self, action: #selector(backTapped(_:)))
         elloNavigationItem.leftBarButtonItems = [item]
@@ -205,14 +205,26 @@ extension CategoryViewController: CategoryStreamDestination, StreamDestination {
 
 extension CategoryViewController: CategoryScreenDelegate {
 
+    public func selectCategoryForSlug(slug: String) {
+        guard let category = categoryForSlug(slug) else { return }
+        selectCatgory(category)
+    }
+
+    private func categoryForSlug(slug: String) -> Category? {
+        return allCategories.filter { $0.slug == slug }.first
+    }
+
     public func categorySelected(index: Int) {
         guard
             let category = allCategories.safeValue(index)
         where category.id != self.category?.id
         else { return }
 
-        Tracker.sharedTracker.categoryOpened(category.slug)
+        selectCatgory(category)
+    }
 
+    public func selectCatgory(category: Category) {
+		Tracker.sharedTracker.categoryOpened(category.slug)
         let streamKind: StreamKind
         switch category.level {
         case .Meta:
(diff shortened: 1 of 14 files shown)


Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -952,6 +952,7 @@ extension StreamViewController: UserDelegate {
 extension StreamViewController: WebLinkDelegate {
 
     public func webLinkTapped(type: ElloURI, data: String) {
+        print(type)
```
colinta: Some debug code left in here?

Sources/Controllers/WebBrowser/ElloWebBrowserViewController.swift
```diff
@@ -164,6 +163,13 @@ extension ElloWebBrowserViewController : WebLinkDelegate {
         }
     }
 
+    private func showCategory(slug: String) {
+        if alreadyOnCategory(slug) { return }
+        let vc = CategoryViewController(slug: slug)
```
colinta: We could add a check here for wether the user is viewing the `CategoryViewController`, and it should be easy to add a func similar to `categorySelected(index: Int)` that accepts a slug (and that would then be called from `categorySelected(index: Int)`), and loads the category "inline"

colinta: (rather than push on another category view controller)

Sources/Controllers/Categories/CategoryViewController.swift
```diff
@@ -28,7 +28,7 @@ public final class CategoryViewController: StreamableViewController {
     }
 
     override public func loadView() {
-        self.title = category?.name
+        self.title = category?.name ?? slug.uppercaseFirst
```
colinta: This feels fishy to me, since "slugs" are not generally human readable.  The contests, for example, would be "Hi-fructose-x-ello"

Sources/Controllers/Categories/CategoryViewController.swift
```diff
@@ -205,12 +205,25 @@ extension CategoryViewController: CategoryStreamDestination, StreamDestination {
 
 extension CategoryViewController: CategoryScreenDelegate {
 
+    public func selectCategoryForSlug(slug: String) {
+        guard let category = categoryForSlug(slug) else { return }
+        selectCatgory(category)
+    }
+
+    private func categoryForSlug(slug: String) -> Category? {
+        return allCategories.filter { $0.slug == slug }.first
```
colinta: We've got an array extension for this, `allCategories.find { $0.slug == slug }`, which also does an "early return" (returns a match as soon as it's found).

steam: 👍

Sources/Controllers/Categories/CategoryViewController.swift
steam: We don't have a `Category` when `loaodView()` is called. We only have a slug. So we can wait to show a title until the `Category` is loaded.

colinta: `name` is always available on `Category` is, but `category` is not available at `loadView`-time.

steam: @rynbyjn - `category` is optional here

colinta: @rynbyjn Since `category` is optional, we have to guard against whether it will be loaded or not.

steam: @colinta - slug is all we've got at first. We can delay it but then we'll be displaying nothing until the category loads in. I'm not sure which is better. I'm pretty sure we always show a title when pushing a VC onto a nav controller. Thoughts?

rynbyjn: Cool, I guess it is just weird to see a `CategoryViewController` that doesn't need a `category`, but I guess if they aren't loaded yet this makes sense..

steam: It originally needed one but if you deep link into the app it may not have it so we're able to show the screen prior to loading one. We immediately load one when `CategoryViewController` is displayed.

colinta: There are a lot of places where we load the `CategoryViewController` and have a `Category` instance - maybe we pass in the title from there?  I'd rather pass it in explicitly if we have it than display something that isn't meant to be displayed.

rynbyjn: Gotcha, makes sense I'll go back to my javascript world and leave y'all alone 😜

steam: I wonder if we let `CategoryGenerator` look it up in the store. We've most likely got it in there whenever we have it to pass in. That keeps it clean, prevents the two paths through `CategoryViewController` which has been nice.

colinta: We'll need the `category.id` in that case... I was picturing an expanded `init`, and still one code-path:
```swift
let controller = CategoryViewController(slug: category.slug, name: category.name)
// ->
init(slug: String, name: String? = nil) {
  self.title = name
  ...
}
```

steam: Good point. Too bad we don't have a real persistence layer that we can query based off different properties. I like the expanded init route for sure 👏.

steam:
Executed 2382 tests, with 0 failures (0 unexpected) in 293.849 (295.328) seconds

PR #218, opened 2016-11-15
Category Header Tweaks

steam:
Primarily, this streamlines the handling of deeplinks as well as enabling deep linking onto the top of a new navigation controller stack (in ProfileCategories).

### DIFF
--- a/Sources/Controllers/App/AppViewController.swift
+++ b/Sources/Controllers/App/AppViewController.swift
@@ -451,25 +451,24 @@ extension AppViewController {
         case .ExploreRecommended,
              .ExploreRecent,
              .ExploreTrending:
-            showDiscoverScreen()
+            showDiscoverScreen(vc)
         case .Discover:
-            showDiscoverScreen()
+            showDiscoverScreen(vc)
         case .DiscoverRandom,
              .DiscoverRecent,
              .DiscoverRelated,
              .DiscoverTrending,
              .Category:
-            showCategoryScreen(data)
+            showCategoryScreen(vc, slug: data)
         case .Invitations:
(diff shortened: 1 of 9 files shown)


Sources/Controllers/App/AppViewController.swift
```diff
@@ -677,12 +686,37 @@ extension AppViewController {
...
         ElloWebBrowserViewController.elloTabBarController?.selectedTab = tab
     }
+
+
+}
+
+extension AppViewController {
+
+    func topViewController(base: UIViewController?) -> UIViewController? {
+        if let nav = base as? UINavigationController {
```
colinta: Nice!

Sources/Utilities/DeepLinking.swift
```diff
@@ -0,0 +1,89 @@
+////
+///  DeepLinking.swift
+//
+
+public struct DeepLinking {
+
```
colinta: I think some things here would be pretty easy to set up specs for, if you're not too pressed for time.

Sources/Controllers/App/AppViewController.swift
steam: I can't take much credit. Stack Overflow was a big help here. 😉

Sources/Utilities/DeepLinking.swift
steam: Yeah, I had that in mind too. I'm not sure how pressed we are. My hunch is that we're going to end up adding the grid/list toggle so we probably have the breathing room.

steam:
Added specs, gonna merge.

PR #222, opened 2016-11-17
Fixes inability to tap on categories in list nav

steam:
* Tapping on categories in the list nav was failing due to a local variable shadowing the class instance variable with the same name. 

* Selecting "Featured" failed on a force unrwap when creating a `DiscoverType`.


colinta:
I can't change tags from my phone.

steam:
Thanks for the review. :) I changed it for you.

PR #224, opened 2016-11-18
Refactor the 'updateInsets' helper to always set 'topInset' according to the 'nav bar' (which is often a button container)

colinta:
update Search, Profile, and Category with new logic.


colinta:
This needs some spec fixes.

Sources/Controllers/Search/SearchScreen.swift
```diff
@@ -132,12 +148,15 @@ public class SearchScreen: UIView, SearchScreenProtocol {
         if let gridListItem = gridListItem where hasGridViewToggle {
             navigationItem.rightBarButtonItems = [gridListItem]
         }
+        else {
+            navigationItem.rightBarButtonItems = []
```
steam: 👀

PR #231, opened 2016-11-28
Stream Zero States

steam:
Adds a adjustable height "Nothing To See Here" cell. Additionally, adds specs for `StreamCellType`.

[Finishes #130410907](https://www.pivotaltracker.com/story/show/130410907)
emptystreamcell__iphone_5_should_match_snapshot 2x


steam:
Executed 2441 tests, with 0 failures (0 unexpected) in 2398.232 (2400.145) seconds

PR #236, opened 2016-11-29
Adds a user's location to their profile

steam:
pasted image at 2016_11_29 03_23 pm


Sources/Controllers/Profile/ProfileHeaderLayoutView.swift
```diff
@@ -13,13 +13,17 @@ public class ProfileHeaderLayoutView: ProfileBaseView {
...
+            statsView.grayLineVisible = bioLinksOrLocationHaveContent
+
+            let locationOrLinksHasContent = calculatedCellHeights.profileLocation > 0 || calculatedCellHeights.profileLinks > 0
+            bioView.grayLineVisible = locationOrLinksHasContent
 
             let linksHasContent = calculatedCellHeights.profileLinks > 0
-            bioView.grayLineVisible = linksHasContent
+            locationView.grayLineVisible = linksHasContent
 
             setNeedsLayout()
```
colinta: How hard would it be to have snapshot views for all these different scenarios?

steam: Good idea, I'll plug away on this in between meetings today.

PR #237, opened 2016-11-29
Adds AutoCompleteViewController to Settings when Location changes

colinta:
When location results come in, they're "coerced" into `AutoCompleteResult` objects by setting `name` to the `location` value.  Not terribly intuitive (?) but maximizes code reuse.

### DIFF
--- a/Sources/Controllers/Settings/SettingsViewController.swift
+++ b/Sources/Controllers/Settings/SettingsViewController.swift
@@ -200,6 +200,7 @@ public class SettingsViewController: UITableViewController, ControllerThatMightH
         else {
             autoCompleteVC.view.frame.size.height = 0
         }
+        updateAutoComplete()
     }
 
     private func updateCurrentUser(user: User) {
--- a/Sources/Controllers/Settings/SettingsViewController.swift
+++ b/Sources/Controllers/Settings/SettingsViewController.swift
@@ -262,13 +262,14 @@ public class SettingsViewController: UITableViewController, ControllerThatMightH
         nameTextFieldView.label.setLabelText(InterfaceString.Settings.Name)
         nameTextFieldView.textField.text = currentUser?.name
 
-        let updateNameFunction = debounce(0.5) { [unowned self] in
-            let name = self.nameTextFieldView.textField.text ?? ""
+        let updateNameFunction = debounce(0.5) { [weak self] in
+            guard let sself = self else { return }
+            let name = sself.nameTextFieldView.textField.text ?? ""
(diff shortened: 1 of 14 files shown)


Sources/Controllers/Settings/SettingsViewController.swift
```diff
@@ -458,3 +531,21 @@ extension SettingsViewController {
     }
 
 }
+
+extension SettingsViewController: AutoCompleteDelegate {
+    public func updateAutoComplete() {
+        self.autoCompleteVC.view.alpha = (locationTextViewSelected && locationAutoCompleteResultCount > 0) ? 1 : 0
```
steam: Super minor but we don't need the `self` here. It's not used a few lines down.

Sources/Model/AutoCompleteResult.swift
```diff
@@ -45,7 +45,8 @@ public final class AutoCompleteResult: JSONAble {
     override public class func fromJSON(data: [String: AnyObject]) -> JSONAble {
         let json = JSON(data)
         Crashlytics.sharedInstance().setObjectValue(json.rawString(), forKey: CrashlyticsKey.AutoCompleteResultFromJSON.rawValue)
-        let result = AutoCompleteResult(name: json["name"].string)
+        let name = json["name"].string ?? json["location"].string
```
steam: Seems fine. I agree that reuse outweighs strict correctness from a naming perspective here.

Sources/Controllers/Settings/SettingsViewController.swift
```diff
@@ -273,19 +325,39 @@ public class SettingsViewController: UITableViewController, ControllerThatMightH
...
+        locationTextFieldView.label.setLabelText(InterfaceString.Settings.Location)
+        locationTextFieldView.textField.keyboardAppearance = .Dark
+        if let user = currentUser, location = user.location {
+            locationTextFieldView.textField.text = location
+        }
 
-        bioTextViewDidChange = debounce(0.5) { [unowned self] in
-            let bio = self.bioTextView.text
-            ProfileService().updateUserProfile(["unsanitized_short_bio": bio], success: { user in
+        let updateLocationFunction = debounce(0.5) { [unowned self] in
```
colinta: Yeah I think you're right, that `weak` is the way to go here.

`unowned` implies that the lifetime of the closure is *directly* tied to the owner-object.  In this case, though, there's a time-lag (0.5s) so that the closure can exist outside the lifetime of the owner --> which implies `weak`.

PR #239, opened 2016-12-01
Show the ProfileTotalCountView as soon as 'total_views_count' is available from the API

colinta:
Show/hide the `badgeButton`, too, either in the avatar view or total count view.

The behavior is a bit odd if the `total_views_count` is not always present on the user, since we don't load users if they've already been loaded.


Sources/Controllers/Profile/Views/ProfileTotalCountView.swift
```diff
@@ -98,22 +89,29 @@ extension ProfileTotalCountView {
...
 
-        let count = ElloAttributedString.style(countText + " ", [
-            NSForegroundColorAttributeName: UIColor.blackColor(),
-            NSParagraphStyleAttributeName : paragraphStyle
-        ])
-        attributed.appendAttributedString(count)
-        attributed.appendAttributedString(totalViewsText)
-        return attributed
-    }
+        if let count = count where count != "" {
```
steam: We could use `count.isEmpty` here.

Sources/Model/User.swift
```diff
@@ -52,6 +52,14 @@ public final class User: JSONAble {
     public var backgroundPosition: String?
     public var onboardingVersion: Int?
     public var totalViewsCount: Int?
+    public var formattedTotalCount: String? {
```
steam: This should probably be a function. :trollface:

jayzes:
Does this require that we push that null attribute on the API response before it goes to the store?

colinta:
@jayzes No, we don't need anything - it looks for an int, but defaults to `nil`.

colinta:
@jayzes @Justin-Holmes btw we're looking for an int in the json called `total_views_count`

jayzes:
@colinta FWIW, wonder if we should make it a `double` or a `long` (not sure which swift has) just to future-proof it

colinta:
@jayzes The default `Int` type is pretty robust, it's not the C `int` type of yesteryear:

```
> Int.max
$R0: Int = 9223372036854775807
```

jayzes:
@colinta excellent!

PR #240, opened 2016-12-01
Fixes onboarding links validation

colinta:
Links are now validated on the client, and the user cannot continue if the link field is invalid, even if they entered other info.

Also discovered a bug where an existing user's links would not be displayed by the app if they deleted their bio.  The changes *are* saved, though, and it's only when updatind an existing user (would not affect onboarding).

### DIFF
--- a/Sources/Controllers/Onboarding/CreateProfileViewController.swift
+++ b/Sources/Controllers/Onboarding/CreateProfileViewController.swift
@@ -13,14 +13,16 @@ public class CreateProfileViewController: UIViewController, HasAppController {
     var didSetName = false
     var didSetBio = false
     var didSetLinks = false
+    var linksAreValid = false
+    var debouncedLinksValidator = debounce(0.5)
     var didUploadCoverImage = false
     var didUploadAvatarImage = false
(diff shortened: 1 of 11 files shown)


Sources/Controllers/Onboarding/CreateProfileViewController.swift
```diff
@@ -39,45 +41,73 @@ extension CreateProfileViewController: CreateProfileDelegate {
...
+            else {
+                didSetLinks = true
+            }
+            linksAreValid = false
+        }
+        onboardingViewController?.canGoNext = profileIsValid
+
+        debouncedLinksValidator { [weak self] in
+            guard let sself = self else { return }
+            sself.screen.linksValid = sself.didSetLinks ? (sself.linksAreValid ? true : false) : nil
```
steam: Can this be simplified to something like this? 
`sself.screen.linksValid = sself.didSetLinks ? sself.linksAreValid : nil`

colinta: Oh yeah, duh, I don't know why I didn't do that...

PR #242, opened 2016-12-08
Announcement push notifications

steam:
This PR does not yet handle loading in announcement notifications. We'll need to add that in once the announcement stuff is baked.


jayzes:
Do we need to add the setting to the store/schema too?

jayzes:
(cc @justin-holmes)

steam:
Great catch @jayzes, we definitely do.

steam:
@jayzes / @Justin-Holmes - Is the announcement push notification PR up on any staging servers for me to verify? I'm checking for the `notify_of_annnouncements_via_push` property in https://github.com/ello/ello/pull/1897/files#diff-39382771bce66dbb2411e857dfe2c7be

Justin-Holmes:
@steam Pushing that up at to stage2 right now, so it should be up in a few.

Specs/Networking/ElloURISpec.swift
```diff
@@ -179,6 +180,7 @@ class ElloURISpec: QuickSpec {
                     }
                 }
 
+"notifications/all"
```
colinta: This is a funny little string. Hey string! How are ya? :-D

steam:
Turns out none of this is needed. We changed the structure of the announcement push notifications server side. 🎉 

No code is the best code.

steam:
**_JUUUUUUUST Kidding_**. We do need the settings for announcement push notifications. Thankfully @jayzes remembers this, over and over again. Because I certainly don't for some reason. The PR **_should_** be good to go now.

jayzes:
~No~Some code is the best code.

PR #249, opened 2016-12-15
Adds Announcement feature

colinta:
Converts NotificationsViewController to use a Generator, and loads Announcements on demand (when loading the notifications tab) and when `NewContentService` indicates new announcements are available.

Tapping on the announcement opens the URL (either as a detail or in-app browser), tapping the "X" dismisses it, posting a PATCH to the API.


Sources/Controllers/Notifications/NotificationCell.swift
```diff
@@ -58,16 +59,16 @@ public class NotificationCell: UICollectionViewCell, UIWebViewDelegate {
...
 
-    var avatarButton: AvatarButton!
-    var buyButtonImage: UIImageView!
-    var replyButton: StyledButton!
-    var relationshipControl: RelationshipControl!
-    var titleTextView: ElloTextView!
-    var createdAtLabel: UILabel!
-    var messageWebView: UIWebView!
-    var notificationImageView: FLAnimatedImageView!
+    let avatarButton = AvatarButton()
```
steam: Nice refactor pulling these up here.

Sources/Controllers/Notifications/NotificationsGenerator.swift
```diff
@@ -0,0 +1,130 @@
+////
+///  NotificationsGenerator.swift
+//
+
+public final class NotificationsGenerator: StreamGenerator {
+    public var currentUser: User?
+    public var streamKind: StreamKind
+
+    private var pinnedNotifications: [AnyObject]?
```
steam: It doesn't look like `pinnedNotifications` is referenced by anything.

Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -1073,18 +1096,23 @@ extension StreamViewController: UICollectionViewDelegate {
...
             user = notification.subject as? User
         {
             userTapped(user)
         }
+        else if let announcement = dataSource.jsonableForIndexPath(indexPath) as? Announcement,
+            callToAction = announcement.ctaURL
+        {
+            Tracker.sharedTracker.announcementOpened(announcement)
+            let request = NSURLRequest(URL: callToAction)
+            ElloWebViewHelper.handleRequest(request, webLinkDelegate: self)
```
steam: What if the announcement's `ctaURL` is an internal deep link?

Sources/Controllers/Stream/StreamKind.swift
```diff
@@ -195,13 +199,14 @@ public enum StreamKind {
                     return accum
                 }
             }
-            else if let comments = jsonables as? [ElloComment] {
-                return comments
+            else if let jsonables = jsonables as? [ElloComment] {
```
steam: These look odd now that they all do the same thing. I'm not sure if it is worth enumerating an array of types to check against or if this is more clear. I like the naming refactor, it just illustrates the repetitiveness of some of these checks.

Sources/Networking/NewContentService.swift
```diff
@@ -87,6 +88,22 @@ private extension NewContentService {
...
     }
 
+    func checkForNewAnnouncements(done: BasicBlock = {}) {
+        let storedKey = StreamKind.Announcements.lastViewedCreatedAtKey
+        let storedDate = GroupDefaults[storedKey].date
+
+         ElloProvider.shared.elloRequest(
+             ElloAPI.AnnouncementsNewContent(createdAt: storedDate),
+             success: { (_, responseConfig) in
+                 if let statusCode = responseConfig.statusCode {// where statusCode == 204
```
steam: Should `// where statusCode == 204` be commented out here?

colinta: Oh shoot, missed this!

Sources/Controllers/Notifications/NotificationsGenerator.swift
colinta: Good catch - these became `announcements`

Sources/Controllers/Stream/StreamViewController.swift
colinta: Isn't that handled by the `webLinkDelegate` arg?  I copied this from `StreamTextCell`, so it *should* handle both internal and external links.

Sources/Controllers/Stream/StreamKind.swift
colinta: Yeah, not sure what the correct refactor is...

could be
```swift
else if let jsonables = (jsonables as? [A] ?? jsonables as? [B] ?? ...)
```

Sources/Controllers/Stream/StreamViewController.swift
steam: Ah, yep. You're right.

Sources/Controllers/Stream/StreamKind.swift
steam: A bit better but I'm guessing the compiler would take longer. 😄 Thanks Trump.

PR #250, opened 2016-12-15
Tweaks to the new zero states

steam:
1. Remove legacy following/starred no results text
1. Reassign empty stream cell on grid/list toggle

### DIFF
--- a/Sources/Controllers/Stream/StreamContainerViewController.swift
+++ b/Sources/Controllers/Stream/StreamContainerViewController.swift
@@ -193,7 +193,6 @@ public class StreamContainerViewController: StreamableViewController {
             self.addChildViewController(vc)
             vc.didMoveToParentViewController(self)
             ElloHUD.showLoadingHudInView(vc.view)
-
         }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/Controllers/Stream/StreamContainerViewController.swift
```diff
@@ -194,19 +194,6 @@ public class StreamContainerViewController: StreamableViewController {
             vc.didMoveToParentViewController(self)
             ElloHUD.showLoadingHudInView(vc.view)
 
-            switch kind {
-            case .Following:
-                let noResultsTitle = InterfaceString.FollowingStream.NoResultsTitle
-                let noResultsBody = InterfaceString.FollowingStream.NoResultsBody
-                friendsViewController = vc
```
steam: Hmm, I wonder what line 201 is about.

colinta: That is the only place the controller is assigned - tread lightly, I think that would break line 273

colinta: That is the only place the controller is assigned - tread lightly, I think that would break line 273

colinta: (but I don't know why `friendsVC` is assigned - and not similarly for `noiseVC`)

PR #255, opened 2017-01-09
Fixes internal ninja links in announcement push notifications

steam:
[fixes #134104237](https://www.pivotaltracker.com/story/show/134104237)


steam:
The specs were failing compilation on those two cells. Adding public got a compile.

PR #258, opened 2017-01-09
ON HOLD DON'T MERGE - Adds Google Conversion Tracking to App

steam:
In order to know the effectiveness of ads we purchase we need to know where a download and install of the iOS app came from. [Google App Conversion Tracking](https://developers.google.com/app-conversion-tracking/ios/?hl=en_US) helps us connects the dots.

Both the `conversionId` and `conversionLabel` are stored in CocoaPods Keys similarly to other pieces of sensitive data.

We'll be updating our privacy policy with this additional information and will link to it here when it is ready.

The addition of GCT requires inclusion of Apple's AdSupport framework.

From Google's App Conversion Tracking documentation:
> The Google Conversion Tracking SDK utilizes Apple's advertising identifier introduced in iOS 6 (IDFA). The SDK uses IDFA under the guidelines laid out in the iOS developer program license agreement. Developers must ensure they are in compliance with the iOS developer program license agreement policies governing the use of this identifier.


Sources/AppDelegate.swift
```diff
@@ -58,6 +58,16 @@ public class AppDelegate: UIResponder, UIApplicationDelegate {
             PushNotificationController.sharedController.receivedNotification(application, userInfo: payload)
         }
 
+#if DEBUG
+        print("Skipping Conversion Tracking in DEBUG")
```
colinta: I dig the "don't track this in the sim" code here :smiley:

Podfile
```diff
@@ -66,6 +66,7 @@ end
 target 'Ello' do
   common_pods
   ello_app_pods
+  pod 'GoogleConversionTracking'
```
steam: This is a bit odd. The specs wouldn't compile with this up in ello_app_pods.

Sources/AppDelegate.swift
```diff
@@ -58,6 +58,16 @@ public class AppDelegate: UIResponder, UIApplicationDelegate {
...
+#else
+        ACTConversionReporter.reportWithConversionID(
```
steam: @rmthamer, this is the line of code that we are calling. It is the only place we are calling anything related to the conversion reporter. The function it is in is called every time the app launches on a device that it is't currently running on. The `conversionID` and `conversionLabel` are the values you guys provided us. Please note that this repository is public so don't type anything sensitive like keys or secrets here. On our call you mentioned pointing out our authentication code. I'm happy to point it out but we're not interacting with any of it for this feature since the conversion tracking happens immediately when the app launches, prior to logging the user into an active session. cc'ing @caseydoran so he is in the loop.

rmthamer: This looks good from my side.  I'm still not clear as to how the attribution works but from the api side this code uses 100% of what is made available so we'll have to see if google is doing something interesting on the backend.  We'll be sending some stuff regarding the testing prior to sending up to the app store which should be helpful.

colinta:
Should we just close this, and open a new PR if we need to?

steam:
Sounds good to me @colinta. Closing.

PR #260, opened 2017-01-10
Prevent double tap loves when disabled

### LINKED ISSUE #257: Hashing contacts
Hey guys,

Was just looking through the codebase and sniffing some of the network traffic to learn more about how Ello utilizes the address book and contacts to make inviting and connecting with friends easier. Have some quick questions for you based on what I found.

First question is, why did you guys decide to only use email addresses? Is there a reason that you decided to stay away from phone numbers? I'm not looking for a ton of info, just curious what thought process went into this.

Second, where is the actual hashing of the emails taking place? I've been digging through the code but can't seem to pinpoint where the email addresses actually get hashed. Last but not least, what are you using to perform the hashing? I assumed you were using something like a port of `bcrypt` but after taking a quick look at your pods I couldn't find anything.

I'm curious about all of this because I'll be implementing similar functionality soon and have always been a big fan of this codebase and your work, so would love to hear any insight you could provide.

Thanks.

steam:
Double tapping the post of a user with loves disabled is no longer allowed.
[Fixes #257 133266763](https://www.pivotaltracker.com/story/show/133266763)

### DIFF
--- a/Sources/Controllers/Stream/StreamViewController.swift
+++ b/Sources/Controllers/Stream/StreamViewController.swift
@@ -807,6 +807,9 @@ extension StreamViewController: StreamEditingDelegate {
             let post = dataSource.postForIndexPath(path),
             let footerPath = dataSource.footerIndexPathForPost(post)
         {
+            guard let lovesEnabled = post.author?.hasLovesEnabled,
+                lovesEnabled == true else { return }
+
             if let window = cell.window {
                 let fullDuration: TimeInterval = 0.4
                 let halfDuration: TimeInterval = fullDuration / 2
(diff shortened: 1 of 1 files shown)


Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -807,6 +807,9 @@ extension StreamViewController: StreamEditingDelegate {
             let post = dataSource.postForIndexPath(path),
             let footerPath = dataSource.footerIndexPathForPost(post)
         {
+            guard let lovesEnabled = post.author?.hasLovesEnabled,
+                lovesEnabled == true else { return }
```
colinta: If you don't need the `lovesEnabled` value, the one liner `post.author?.hasLovesEnabled == true` works here, too.

steam: Good catch 👍

PR #261, opened 2017-01-11
Fixes Announcement tracking

colinta:
The tracking code was in the placeholders callback, not replace-items, and needed to be triggered on `viewDidAppear`

### DIFF
--- a/Sources/Controllers/Notifications/NotificationsViewController.swift
+++ b/Sources/Controllers/Notifications/NotificationsViewController.swift
@@ -266,6 +266,5 @@ extension NotificationsViewController: AnnouncementDelegate {
         for announcement in announcements {
             Tracker.sharedTracker.announcementViewed(announcement)
         }
-        print("=============== \(#file) line \(#line) ===============")
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/Controllers/Notifications/NotificationsViewController.swift
```diff
@@ -77,6 +77,13 @@ public class NotificationsViewController: StreamableViewController, Notification
         PushNotificationController.sharedController.updateBadgeNumber(0)
     }
 
+    override public func viewDidAppear(animated: Bool) {
+        super.viewDidAppear(animated)
+
+        let jsonables = streamViewController.dataSource.streamCellItems.map { $0.jsonable }
```
steam: This seems generally handy. I wonder if we should add a `jsonables` property on `StreamDataSource` for all to use.

colinta: Surprisingly I couldn't find another place that would share this bit of code - though I *would* like to decrease the "reach through" nature of this code.

PR #266, opened 2017-01-12
Remove autocomplete jitter

steam:
When auto completing usernames and emojis the suggestions were jumping around due to addition/removal of the native iOS spelling suggestions. 

The spelling suggestions should remain hidden until the emoji/username suggestions are gone.

Removing `eagerCheck` and re-assigning `textView.autocorrectionType = .yes` in `hideAutoComplete(_ textView: UITextView)` seem to do the trick. Any flaws in my thinking @colinta?

[fixes #137193245](https://www.pivotaltracker.com/story/show/137193245)


steam:
@colinta - I re-added `eagerCheck` using the `wordStartIndex` changes you suggested. Nice find 🤘

PR #267, opened 2017-01-12
Changes 'sself' and 'weakSelf' and 'wSelf' to `self`/self

colinta:
Conversion was a PITA, but *yay* swift 3 syntax.


colinta:
wakka wakka wakka!

PR #268, opened 2017-01-13
Locks down dependencies

steam:
A few dependencies were not locked down tightly enough. 

`ImagePickerSheetController` removed their `swift2.3-improvements` branch out from under us and `TimeAgoInWords` and `SVGKit` changed. `Podfile` now locks down to individual commits on those 3 pods.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -93,7 +93,7 @@ end
 ######################### PRIVATE LANES #########################
 
 private_lane :switch_to_prod do
-  `bundle exec rake generate:staging_keys`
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -93,7 +93,7 @@ end
 ######################### PRIVATE LANES #########################
 
 private_lane :switch_to_prod do
-  `bundle exec rake generate:staging_keys`
+  `bundle exec rake generate:prod_keys`
```
jayzes: :trollface:

PR #271, opened 2017-01-13
Merges recent commits from master into 1.20

colinta:


colinta:
@steam These didn't make it into 1.20, we'll need to send out another TestFlight release.

colinta:
I'm including `sd/1.20/build-5558`, just to be sure there's no conflicts (there weren't!)

PR #275, opened 2017-01-16
Fixes to Notifications and other Stream loading/reloading

colinta:
Most of the logic changes are in `NotificationsViewController`, but all the Generator-based controllers were clearing the HUD in `setPlaceholders`, which meant the HUD was going away immediately since that callback is called immediately.

The `ElloCollectionView` code is being reverted, we already reverted it in swift3.0, it was meant to fix a minor bug, but causes an issue with pull-to-refresh.

### DIFF
--- a/Sources/Controllers/Categories/CategoryViewController.swift
+++ b/Sources/Controllers/Categories/CategoryViewController.swift
@@ -151,7 +151,7 @@ extension CategoryViewController: CategoryStreamDestination, StreamDestination {
 
     public func setPlaceholders(items: [StreamCellItem]) {
         streamViewController.clearForInitialLoad()
-        streamViewController.appendUnsizedCellItems(items, withWidth: view.frame.width) { _ in }
+        streamViewController.appendStreamCellItems(items)
     }
 
     public func setPrimaryJSONAble(jsonable: JSONAble) {
@@ -171,6 +171,7 @@ extension CategoryViewController: CategoryStreamDestination, StreamDestination {
             self.pagePromotional = pagePromotional
         }
         updateInsets()
+        streamViewController.doneLoading()
     }
 
     public func setCategories(categories: [Category]) {
--- a/Sources/Controllers/Notifications/NotificationsGenerator.swift
+++ b/Sources/Controllers/Notifications/NotificationsGenerator.swift
@@ -33,8 +33,10 @@ public final class NotificationsGenerator: StreamGenerator {
             announcements = []
             notifications = []
         }
+        else {
+            setPlaceHolders()
+        }
 
-        setPlaceHolders()
         loadAnnouncements()
         loadNotifications()
     }
--- a/Sources/Controllers/Notifications/NotificationsViewController.swift
+++ b/Sources/Controllers/Notifications/NotificationsViewController.swift
@@ -63,7 +63,7 @@ public class NotificationsViewController: StreamableViewController, Notification
         scrollLogic.prevOffset = streamViewController.collectionView.contentOffset
         scrollLogic.navBarHeight = 44
 
-        reload()
+        initialLoad()
     }
 
     override public func viewWillAppear(animated: Bool) {
@@ -93,7 +93,6 @@ public class NotificationsViewController: StreamableViewController, Notification
         ElloHUD.showLoadingHudInView(streamViewController.view)
         hasNewContent = false
 
-        generator?.streamKind = categoryStreamKind
         generator?.load(reload: true)
     }
 
@@ -127,7 +126,6 @@ public class NotificationsViewController: StreamableViewController, Notification
         updateInsets()
     }
 
-
     // used to provide StreamableViewController access to the container it then
     // loads the StreamViewController's content into
     override func viewForStream() -> UIView {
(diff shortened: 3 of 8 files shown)


Sources/Controllers/Categories/CategoryViewController.swift
```diff
@@ -151,7 +151,7 @@ extension CategoryViewController: CategoryStreamDestination, StreamDestination {
 
     public func setPlaceholders(items: [StreamCellItem]) {
         streamViewController.clearForInitialLoad()
-        streamViewController.appendUnsizedCellItems(items, withWidth: view.frame.width) { _ in }
+        streamViewController.appendStreamCellItems(items)
```
colinta: calling `appendUnsizedCellItems` has the side effect that `StreamViewController` hides the HUD, so these were all changed to `appendStreamCellItems`.

Sources/Controllers/Categories/CategoryViewController.swift
```diff
@@ -171,6 +171,7 @@ extension CategoryViewController: CategoryStreamDestination, StreamDestination {
             self.pagePromotional = pagePromotional
         }
         updateInsets()
+        streamViewController.doneLoading()
```
colinta: But now, since we aren't relying on `appendUnsizedCellItems` to hide the HUD, we need to tell the `StreamViewController` that the content is ready / hide the HUD.

Sources/Controllers/Notifications/NotificationsGenerator.swift
```diff
@@ -33,8 +33,10 @@ public final class NotificationsGenerator: StreamGenerator {
             announcements = []
             notifications = []
         }
+        else {
+            setPlaceHolders()
+        }
```
colinta: This is a departure from the usual Generator code - after a reload we will call `replacePlaceholders`, so it's not necessary to reset the placeholder items, which also clears the stream cell items.

Sources/Controllers/Notifications/NotificationsViewController.swift
```diff
@@ -93,7 +93,6 @@ public class NotificationsViewController: StreamableViewController, Notification
         ElloHUD.showLoadingHudInView(streamViewController.view)
         hasNewContent = false
 
-        generator?.streamKind = categoryStreamKind
```
colinta: This is set in the `activatedCategory` method instead.

Sources/Controllers/Stream/Detail/PostDetailViewController.swift
```diff
@@ -222,26 +222,28 @@ extension PostDetailViewController: StreamDestination {
...
 
     public func replacePlaceholder(type: StreamCellType.PlaceholderType, items: [StreamCellItem], completion: ElloEmptyCompletion) {
-        streamViewController.replacePlaceholder(type, with: items, completion: completion)
-    }
-
-    public func setPlaceholders(items: [StreamCellItem]) {
-        streamViewController.clearForInitialLoad()
-        streamViewController.appendUnsizedCellItems(items, withWidth: view.frame.width) { _ in
+        streamViewController.replacePlaceholder(type, with: items) {
             if let scrollToComment = self.scrollToComment {
```
colinta: This `scrollToComment` check was in `setPlaceholders`, which meant that it would never *find* the comment that was supposed to be scrolled to.

Sources/Controllers/Stream/Detail/PostDetailViewController.swift
```diff
@@ -222,26 +229,23 @@ extension PostDetailViewController: StreamDestination {
...
+                self.checkScrollToComment()
+            }
+            completion()
+        }
     }
 
     public func setPlaceholders(items: [StreamCellItem]) {
         streamViewController.clearForInitialLoad()
-        streamViewController.appendUnsizedCellItems(items, withWidth: view.frame.width) { _ in
-            if let scrollToComment = self.scrollToComment {
```
colinta: This scrollToComment check was in setPlaceholders, which meant that it would never find the comment that was supposed to be scrolled to.

Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -313,8 +313,8 @@ public final class StreamViewController: BaseElloViewController {
         reloadCells(now: true)
     }
 
-    public func appendUnsizedCellItems(items: [StreamCellItem], withWidth: CGFloat?, completion: StreamDataSource.StreamContentReady? = nil) {
-        let width = withWidth ?? self.view.frame.width
+    public func appendUnsizedCellItems(items: [StreamCellItem], completion: StreamDataSource.StreamContentReady? = nil) {
+        let width = view.frame.width
```
rynbyjn: It looks like we were sometimes passing `nil` and sometimes passing `view.frame.width` to the `withWidth` param before.. Now that it is always the `view.frame.width`, could this potentially mess with the grid layout?

colinta:
```
Test Suite 'All tests' failed at 2017-01-16 12:34:00.880.
	 Executed 2584 tests, with 5 failures (0 unexpected) in 301.003 (302.565) seconds
```

These 5 failures are expected: `AnnouncementCell` snapshots have been updated in swift3.0, and the `AutoCompleteResultSpec` is just an NSCoding version number mismatch.

Sources/Controllers/Stream/StreamViewController.swift
colinta: The places where we passed in `nil` would then *default* to the entire width.  The calculators are responsible for dividing that width up into column widths, e.g.

```swift
if job.columnCount == 1 {
    self.maxWidth = job.width
}
else {
    self.maxWidth = floor(job.width / CGFloat(job.columnCount) - StreamKind.Following.columnSpacing * CGFloat(job.columnCount - 1))
}
```

rynbyjn: Gotcha, word 🤘

colinta:
screen shot 2017-01-16 at 12 47 42 pm

PR #283, opened 2017-01-17
Fixes the double spinner, and another stab at the contentInsets bug

colinta:
Double spinner was showing up on Post Detail, and also on the Category view.

If you tapped a post from notifications and refreshed it, the contentInsets were incorrect, I put in a simple fix for that (work around `SSPullToRefresh` overriding contentInset).

### DIFF
--- a/Sources/Controllers/Stream/StreamViewController.swift
+++ b/Sources/Controllers/Stream/StreamViewController.swift
@@ -208,9 +208,12 @@ public final class StreamViewController: BaseElloViewController {
     public var contentInset: UIEdgeInsets {
         get { return collectionView.contentInset }
         set {
+            // the order here is important, because SSPullToRefresh will try to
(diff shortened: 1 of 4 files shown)


Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -208,9 +208,12 @@ public final class StreamViewController: BaseElloViewController {
     public var contentInset: UIEdgeInsets {
         get { return collectionView.contentInset }
         set {
+            // the order here is important, because SSPullToRefresh will try to
+            // set the contentInset, and that can have weird side effects, so
+            // we need to set the contentInset *after* pullToRefreshView.
+            pullToRefreshView?.defaultContentInset = newValue
```
steam: Nice find 👍

PR #287, opened 2017-01-18
Last minute spinner fixes

colinta:
from lucian


Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -333,6 +333,20 @@ public final class StreamViewController: BaseElloViewController {
         }
     }
 
+    public func hasCellItems(for placeholderType: StreamCellType.PlaceholderType) -> Bool {
```
steam: Should this be in `StreamDataSource`? It looks like all the work is done on info found in their.

colinta: I wondered, I only but it here so that the controllers that use this method don't "reach through" the stream view controller to access the data source... but it makes sense to have it in the data source.

PR #289, opened 2017-01-19
Fix double reposting crash

steam:
We were not preventing a user from reposting the same post twice which caused a crash on the second attempt.

For other actions, such as loving a post, the returned post in the api call has it's `loved` property set to true. Reposting is unique in that the API returns the new repost, not the updated original post. The result was reposted posts found in multiple streams were not updated correctly after reposting. 

This PR updates the reposted `Post` in `ElloLinkedStore` and posts a new experience update, `.reposted`.

[Fixes #288137191093](https://www.pivotaltracker.com/story/show/137191093)

### DIFF
--- a/Sources/Controllers/Stream/StreamDataSource.swift
+++ b/Sources/Controllers/Stream/StreamDataSource.swift
@@ -459,7 +459,7 @@ class StreamDataSource: NSObject, UICollectionViewDataSource {
             }
 
         case .delete:
-            let _ = removeItemsFor(jsonable: jsonable, change: change)
+            removeItemsFor(jsonable: jsonable, change: change)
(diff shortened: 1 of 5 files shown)


Sources/Controllers/Stream/StreamDataSource.swift
```diff
@@ -622,6 +608,7 @@ class StreamDataSource: NSObject, UICollectionViewDataSource {
         collectionView.reloadData()
     }
 
+    @discardableResult
```
colinta: Nice 😃

steam: This guy showed me that the other day. Pretty stoked on it. I'm not sure who he was though. :trollface:

PR #293, opened 2017-01-20
Support commenting on a post via deep-linking when the comment points to a repost that hasn't been loaded

colinta:
Yeah, that.


colinta:
Yeah, and makes a weird problem (finding the "parent post" with all the repost crazyness) into an easy one (any post id works when commenting)

PR #294, opened 2017-01-20
Parses JWT for isStaff flag

steam:
The auth token is a [JWT](https://jwt.io) which contains several properties. The iOS app is interested in the `is_staff` flag. We read that value if it exists and store it in `AuthToken.isStaff`. 

The first use of `AuthToken().isStaff` is found in `AppViewController`. The `DebugController` is now available in both `DEBUG` and non `DEBUG` builds if the authenticated user is a staff member.


colinta:
There's an empty playground here, was that intentional?

steam:
It was intentional. I figured it would be nice to have one in the project now that we can use them. I emptied it out though so its a fresh scratch pad.

PR #296, opened 2017-01-23
Fixes spec compile errors

steam:
Also updates Nimble with the hopes of unclogging Travis

### DIFF
--- a/Specs/Controllers/Omnibar/OmnibarViewControllerSpec.swift
+++ b/Specs/Controllers/Omnibar/OmnibarViewControllerSpec.swift
@@ -80,7 +80,7 @@ class OmnibarViewControllerSpec: QuickSpec {
                     let post = Post.stub([
                         "author": User.stub(["username": "colinta"])
                         ])
-                    subject = OmnibarViewController(parentPost: post)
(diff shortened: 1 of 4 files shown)


Specs/Controllers/Omnibar/OmnibarViewControllerSpec.swift
```diff
@@ -367,7 +367,7 @@ class OmnibarViewControllerSpec: QuickSpec {
                         _ = Tmp.write(data, to: fileName)
                     }
 
-                    subject = OmnibarViewController(parentPost: Post.stub([:]), defaultText: "@666 ")
+                    subject = OmnibarViewController(parentPostId: "123", defaultText: "@666 ")
```
colinta: aw, shoot, thanks for taking care of that.

steam: No worries!

steam:
I've not found anything similar to this. Most people have specific compile errors rather than a hung linking command. People also talk about the simulator not being ready which is why I tried a new command for pre-warming it. Something we were attempting using `xcinstruments`. It didn't help unfortunately. Here's to Nimble 5.1.1 being the magic fix!

PR #297, opened 2017-01-23
Changes behavior of red-dot reloading in Notifications and Streams

colinta:
Tapping red dot in Notifications shows the loader (regression fix), and in Streams tapping the red dot while viewing "Starred" will switch to "Following" and reload the page.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -59,7 +59,7 @@ end
 def spec_pods
   pod 'FBSnapshotTestCase'
   pod 'Quick', '~> 1.0'
-  pod 'Nimble', '5.1'
+  pod 'Nimble', '5.1.1'
   pod 'Nimble-Snapshots', '~> 4.4'
 end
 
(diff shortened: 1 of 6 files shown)


colinta:
Running specs now, btw.  This PR includes `sd/fix/specs`, so we should probably merge that one first.

colinta:
```Executed 2602 tests, with 0 failures (0 unexpected) in 275.870 (276.938) seconds```

jayzes:
@colinta will that reload trigger a reload from the server?

colinta:
@jayzes I hope so!  But yeah, that's the point.  I'm actually not sure I understand your question.

colinta:
Like I'm answering a tautology.  Will telling me your first name give me the information I need to call you by your first name?  *No it will not!*

jayzes:
Haha, fair enough. I was just confirming that it was reloading from the server and not from locally-cached data. Was virtually positive that it was, but wanted to make sure I wasn't missing something.

colinta:
Ahh gotcha - yes, it's a "go fetch the latest data" kinda reload, not the "hey collection view, are you awake?  Can you reload your cells?  No? *Crash!?*  Why did you crash?" kinda reload that is the bread and butter of iOS development. :trollface:

jayzes:
🤘

PR #299, opened 2017-01-24
Ensure only staff and debug builds allow debugging

steam:


colinta:
Ooh, yikes, good catch.

PR #303, opened 2017-01-27
Responders - convert delegates to responders

steam:
Part one of the move away from delegates to utilizing the responder chain with a little added safety.

### DIFF
--- a/Sources/Controllers/Stream/Cells/StreamHeaderCell.swift
+++ b/Sources/Controllers/Stream/Cells/StreamHeaderCell.swift
@@ -62,12 +62,6 @@ class StreamHeaderCell: UICollectionViewCell {
         }
     }
 
-    weak var relationshipDelegate: RelationshipDelegate? {
-        get { return relationshipControl.relationshipDelegate }
-        set { relationshipControl.relationshipDelegate = newValue }
-    }
-
-
     var avatarHeight: CGFloat = 60.0 {
         didSet { setNeedsDisplay() }
     }
--- a/Sources/Controllers/Stream/Cells/StreamHeaderCell.swift
+++ b/Sources/Controllers/Stream/Cells/StreamHeaderCell.swift
@@ -69,7 +69,6 @@ class StreamHeaderCell: UICollectionViewCell {
 
     weak var userDelegate: UserDelegate?
     weak var categoryDelegate: CategoryDelegate?
-    weak var streamEditingDelegate: StreamEditingDelegate?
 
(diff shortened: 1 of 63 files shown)


Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -601,10 +601,13 @@ final class StreamViewController: BaseElloViewController {
         }
     }
 
+    override var next: UIResponder? {
+        return postbarController
```
steam: Adding this allowed `PostbarController` to participate in the responder chain. It might get a little strange when we need more than one custom type (that is not in the UIViewController/UIView responder chain) to respond to events.

Sources/Controllers/Stream/Cells/StreamHeaderCell.swift
```diff
@@ -66,7 +66,7 @@ class StreamHeaderCell: UICollectionViewCell {
         get { return relationshipControl.relationshipDelegate }
         set { relationshipControl.relationshipDelegate = newValue }
     }
-    weak var postbarDelegate: PostbarDelegate?
```
colinta: Nice!  The whole thing just up and gone!

Sources/Controllers/Stream/StreamViewController.swift
colinta: I think we need to return a next-responder from the postbar controller, otherwise the responder chain will end there (the default implementation returns `nil`).  We could kick *back* to the StreamViewController, ala
```
class PostbarController {
  var next: UIResponder? { return streamViewController.nextAfterPostbar }
}
class StreamViewController {
  var nextAfterPostbar: UIResponder? { return super.next }
}
```

steam:
Sounds good. I have the same worry about next responder. I'm planning to do a handful of these and see if anything is busted. The next PRs will merge into this one so that addition is to be determined.

steam:
Sure enough @colinta , the issue of who is the next responder after `PostbarController` is screwed up in `NotificationsViewController`. It does not become the next responder for `NotificationCell` triggered events.

steam:
@colinta - #314 addresses the nextResponder issue you mention above. It is set to merge into this branch. Here is the fix https://github.com/ello/ello-ios/pull/314/commits/598f7f44eaa7ea881833600876c991af126accd6.

steam:
🎉 The responders PRs are all merged in. @colinta, if everything looks good you can approve this PR and I'll merge it into #307.

PR #305, opened 2017-01-27
Implements the LoggedOutViewController

colinta:
- refactored `AppViewController` to use new controller
- added `LoggedOutScreen` and protocols
- removed `IntroController`, because *it was in the damn way*
- removed `StartupController` (unused)

Implementation detail: I moved controllers around so that the view hierarchy is like this:

- `AppViewController`
  - `NavigationController (a)`
    - `LoggedOutController`
      - `NavigationController (b)`
        - `DiscoverAllController`

When showing login and join, they are pushed onto `NavigationController (a)`, all the others go onto `NavigationController (b)`


steam:
-881 🎉

PR #311, opened 2017-01-31
Bust the cache on login

colinta:
@steam I copied `logOutCurrentUser()` and then kept the parts I thought were appropriate (not many).


colinta:
Needs specs on defaults resetting.

colinta:
@steam Specs added!

PR #312, opened 2017-01-31
Truncates text regions even more

colinta:
TextRegion `|>` break up `<p>` tags `|>` break up `<br>` tags `|>` truncate to 7500 characters.

Wondering about the `7500 + 10` code?  It's to remove the very last `</p>`, if it's there (and maybe newlines - this is all a bit hand-wavy).


colinta:
Needs specs on HTML -> [TextRegion] parsing

colinta:
@steam Specs!

PR #313, opened 2017-02-02
Adds and activates the Join and Login buttons in LoggedOutController

colinta:
And some snapshot specs.


colinta:
Needs: close button

colinta:
@steam Close button: added!

PR #316, opened 2017-02-02
Combines inviteResponder and inviteDelegate

steam:
Previously, we had `inviteDelegate` and `inviteResponder` protocols. One was implemented by `StreamableViewController`, the other by `StreamViewController`. Now that we're using the responder chain they can both be moved into `StreamableViewController`. A nice side effect that slims down `StreamViewController` a bit.

### DIFF
--- a/Sources/Controllers/Stream/StreamableViewController.swift
+++ b/Sources/Controllers/Stream/StreamableViewController.swift
@@ -23,7 +23,7 @@ protocol CreatePostDelegate: class {
 @objc
 protocol InviteResponder: class {
     func onInviteFriends()
-    func sendInvite(person: LocalPerson, isOnboarding: Bool, didUpdate: @escaping ElloEmptyCompletion)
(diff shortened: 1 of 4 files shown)


Sources/Controllers/Stream/StreamableViewController.swift
```diff
@@ -21,8 +21,9 @@ protocol CreatePostDelegate: class {
 }
 
 @objc
-protocol InviteResponder: NSObjectProtocol {
+protocol InviteResponder: class {
     func onInviteFriends()
+    func sendInvite(person: LocalPerson, isOnboarding: Bool, didUpdate: @escaping ElloEmptyCompletion)
```
colinta: Nitpicky, but `completion` would be a more standard name for `didUpdate` (esp since it's called on success and failure)

steam: I like the refactor. I'll push up the change and merge. 👍

PR #317, opened 2017-02-03
SimpleStreamDelegate -> SimpleStreamResponder

steam:
This required a bit of work due to `@objc`'s inability to accept swift only types as arguments in `#selector`. The solution is to box swift's value types in a recognizable Objective-C class. A bit ugly but it does the trick.

### DIFF
--- a/Sources/Controllers/Stream/StreamViewController.swift
+++ b/Sources/Controllers/Stream/StreamViewController.swift
@@ -11,8 +11,9 @@ import DeltaCalculator
 
 // MARK: Responder Implementations
 
-protocol SimpleStreamDelegate: class {
-    func showSimpleStream(endpoint: ElloAPI, title: String, noResultsMessages: (title: String, body: String)?)
+@objc
+protocol SimpleStreamResponder: class {
+    func showSimpleStream(boxedEndpoint: BoxedElloAPI, title: String, noResultsMessages: NoResultsMessages?)
 }
 
 @objc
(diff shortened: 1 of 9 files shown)


Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -84,6 +85,17 @@ struct StreamNotification {
...
 }
 
+// This is an NSObject in order to pass it as an
+// objective-c argument to a responder chain call
+class NoResultsMessages: NSObject {
+    let title: String
+    let body: String
+    init(title: String?, body: String?) {
+        self.title = title ?? ""
+        self.body = body ?? ""
```
colinta: Why the `Optional` here?  Are these ever `nil`?

steam: Thats a good question. They stayed optional because of their original usage but really the entire `NoResultsMessages` is optional, not the independent component parts of `title` and `body`. I'll make them required. 👍

steam: Its funny to imagine a `NoResultsMessages` object with two empty strings. An utterly useless object.

Sources/Networking/ElloAPI.swift
```diff
@@ -8,6 +8,13 @@ import Result
 
 typealias MoyaResult = Result<Moya.Response, Moya.Error>
 
+// 😭 I'm as sad as you are about this. We want the responder chain
+// and we want to pass ElloAPI arguments. So we box it.
```
colinta: Haha, well I like it more than the "protocol pair" workaround, so 👍

PR #318, opened 2017-02-03
Implements deep links in logged out views, and refactors OUT link handling from `StreamViewController`

colinta:
:v: Yay to centralized link handling!  Since `AppViewController` was already handling all the deep links, I wondered why we didn't rely on it within `StreamViewController`, too, and couldn't find a good reason.  Posts and Profiles are pushed onto the visible view controller.

### DIFF
--- a/Sources/Controllers/App/AppViewController.swift
+++ b/Sources/Controllers/App/AppViewController.swift
@@ -437,7 +437,10 @@ extension AppViewController {
         Tracker.shared.deepLinkVisited(path)
 
         let (type, data) = ElloURI.match(path)
+        navigateToURI(path: path, type: type, data: data)
+    }
 
+    func navigateToURI(path: String, type: ElloURI, data: String) {
         guard type.shouldLoadInApp else {
             if let pathURL = URL(string: path) {
                 UIApplication.shared.openURL(pathURL)
--- a/Sources/Controllers/App/AppViewController.swift
+++ b/Sources/Controllers/App/AppViewController.swift
@@ -193,7 +193,7 @@ extension AppViewController {
         swapViewController(parentNavController) {}
     }
 
-    func showJoinScreen(animated: Bool, invitationCode: String? = nil) {
+    func showJoinScreen(invitationCode: String? = nil) {
         guard
             let nav = visibleViewController as? UINavigationController,
             let loggedOutController = nav.childViewControllers.first as? LoggedOutViewController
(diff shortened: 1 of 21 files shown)


Sources/Controllers/App/AppViewController.swift
```diff
@@ -702,22 +727,32 @@ extension AppViewController {
...
         }
-        else {
+        else if
+            let nav = self.visibleViewController as? UINavigationController,
+            nav.viewControllers.first is LoggedOutViewController
+        {
+            navController = nav
+        }
+
+        if let navController = navController {
```
steam: This could probably be a simple optional chain `navController?.pushViewController(vc, animated: true)`

Sources/Controllers/Onboarding/CategoriesSelectionViewController.swift
```diff
@@ -79,7 +78,7 @@ extension CategoriesSelectionViewController: OnboardingStepController {
             }
             .onFail { _ in
                 let alertController = AlertViewController(error: InterfaceString.GenericError)
```
steam: What do you think about proactively using `weak self` `strong self` in these?

Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -983,68 +983,13 @@ extension StreamViewController: UserDelegate {
...
-        case .search: DeepLinking.showSearch(navVC: navigationController, currentUser: currentUser, terms: data)
-        case .settings: DeepLinking.showSettings(navVC: navigationController, currentUser: currentUser)
-        }
+    func webLinkTapped(path: String, type: ElloURI, data: String) {
+        guard
+            let parentController = parent as? HasAppController,
+            let appViewController = parentController.appViewController
+        else { return }
+
+        appViewController.navigateToURI(path: path, type: type, data: data)
```
steam: A thing of beauty here @colinta. This seriously made my day 🤘.

Sources/Controllers/Onboarding/CategoriesSelectionViewController.swift
colinta: Yeah I certainly prefer `weak self` these days, are you saying we should try to convert these when we see em? (I didn't think of it here, since the changes were just related to the app controller)

colinta:
I don't have any spec failures in `AnonymousAuthenticationSpec` (though I did just notice some in SearchScreen and AuthState), not sure what needs to change there.

Sources/Controllers/Onboarding/CategoriesSelectionViewController.swift
steam: Exactly, I think we should convert them when we see them. I don't think we'll see them all but I've been changing them as I notice them. And yeah, I wouldn't have assumed you'd think of it. I just happened to notice the opportunity.

Sources/Controllers/App/AppViewController.swift
colinta: K, changed it, though in this case I chose the `if let a = a {}` style because I thought it showed "I may or may not end up with a `navController`, whereas `controller?.doSomething` looks like "I expect controller to be there, but just in case it's not, I don't want an error".

Sources/Controllers/Onboarding/CategoriesSelectionViewController.swift
colinta: Added

colinta:
@steam Ok, got the specs working, I put the `SearchScreen` fixes in their own PR.

PR #321, opened 2017-02-03
CreatePostDelegate -> CreatePostResponder

steam:
This one is nifty for a few reasons. 1) It finally gets rid our long standing "// This is a bit dirty, we should not call a method on a compositionally held controller's createPostDelegate. Can this use the responder chain when we have parameters to pass?" comments. Yay. 2) Rather than assigning a delegate in specs we verify responder behavior by using a mock window that implements the responder functions. Not as elegant but cool that it works.

### DIFF
--- a/Sources/Controllers/Stream/PostbarController.swift
+++ b/Sources/Controllers/Stream/PostbarController.swift
@@ -158,15 +158,13 @@ class PostbarController: UIResponder, PostbarResponder {
     }
 
     func editCommentButtonTapped(_ indexPath: IndexPath) {
-        // This is a bit dirty, we should not call a method on a compositionally held
-        // controller's createPostDelegate. Can this use the responder chain when we have
-        // parameters to pass?
         guard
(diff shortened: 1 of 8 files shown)


Sources/Controllers/Stream/PostbarController.swift
```diff
@@ -335,29 +333,26 @@ class PostbarController: UIResponder, PostbarResponder {
...
-        // parameters to pass?
         guard
             let comment = commentForIndexPath(indexPath),
             let presentingController = presentingController
         else { return }
 
         let postId = comment.loadedFromPostId
-        PostService().loadReplyAll(postId, success: { usernames in
+        PostService().loadReplyAll(postId, success: {[weak self] usernames in
             let usernamesText = usernames.reduce("") { memo, username in
```
colinta: Why not the usual `guard self = self` here?

steam: Cause I'm **LAZY**. 😏 Good catch.

PR #323, opened 2017-02-04
Category delegates -> responders

steam:
These are pretty small so I included both `CategoryResponder` and `SelectedCategoryResponder`. These compile and specs pass but it is unclear to me what causes `func collectionView(_ collectionView: UICollectionView, didDeselectItemAt indexPath: IndexPath)` in `StreamViewController` to be called. I was unable to trigger that callback through actual app use. Any ideas?


colinta:
@steam Not all cells are "selectable" (there's another delegate method to determine that), actually I only recall the onboarding "select 3 categories" items being selectable, so you'll have to trigger onboarding to test that.

steam:
I did try this in onboarding and it never triggered for me. Perhaps it isn't used anymore?

PR #327, opened 2017-02-06
Updates Post Detail and Profile screens for logged-out views

colinta:
Removes the ellipsis button from the navigation bar.

Also noticed that the grid/list toggle was not the same for "meta" categories and "actual" categories, and we had an unused `StreamKind` (`categoryPosts`).

Running specs now.


colinta:
I had some specs to fix, they're all good now, and I added the "background color change" from Lucian, w/ updated snapshot specs.

PR #328, opened 2017-02-07
Just a thought, but on staging I was 'locked out' due to a long API request

colinta:
So this is how we could add a timeout to our API requests.  @caseydoran @steam does this sound like a good idea to you?

### DIFF
--- a/Sources/Controllers/Categories/CategoryGenerator.swift
+++ b/Sources/Controllers/Categories/CategoryGenerator.swift
@@ -141,13 +141,18 @@ private extension CategoryGenerator {
     }
 
     func loadCategory(_ doneOperation: AsyncOperation, reload: Bool = false) {
-        guard !doneOperation.isFinished || reload else { return }
-        guard !usesPagePromo() else { return }
+        guard
+            !doneOperation.isFinished || reload,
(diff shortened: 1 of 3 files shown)


steam:
Seems like a reasonable idea. Is there any way to test out how this feels in the event of a long running request?

caseydoran:
How would we improve the experience for the user? Is this a retry? Right now we show the "something went wrong" alert, right? I think the server times out after 30s, so seems like we should give it all that time. What happened with the "lockout"?

colinta:
I think it was more than 30s, and I'm not entirely sure other than I know it was trying to load a promotion.  Setting it to 30s on the client side makes sense, then, if we know the server isn't going to try any longer than that, and if something is going on w/ the server where the request is getting lost/hung, then we have a client-side fallback.

colinta:
@steam I investigated further and found that it was actually because the response came back with no page promotionals, not a timeout issue.  So I fixed that, but I still like the idea of a client-side timeout (I upped it to 30s per Casey's comment).  Thoughts?

steam:
The timeout seems fine to me. Matching the server timeout makes sense to me.

PR #330, opened 2017-02-08
Relationship Responders

steam:
This was the most intricate of the conversions. The interplay between `RelationshipController`, `RelationshipControl` and their delegate/responders required a lot of Swift value type wrapping and code refactoring. `RelationshiopController` is now held by `BaseElloViewController` due to `BlockModalViewController`'s need for it to be in the responder chain.

Future work may include renaming the `RelationshipPriority` and `RelationshipRequestStatus` wrappers for more clarity.

### DIFF
--- a/Sources/Controllers/BaseElloViewController.swift
+++ b/Sources/Controllers/BaseElloViewController.swift
@@ -33,9 +33,36 @@ class BaseElloViewController: UIViewController, HasAppController, ControllerThat
         return findViewController { vc in vc is BottomBarController } as? BottomBarController
     }
 
+    // This is an odd one, `super.next` is not accessible in a closure that
+    // captures self so we stuff it in a computed variable
+    var superNext: UIResponder? {
+        return super.next
+    }
+
+    var relationshipController: RelationshipController?
+
+    override var next: UIResponder? {
+        return relationshipController
+    }
+
     override func viewDidLoad() {
         super.viewDidLoad()
         self.navigationItem.fixNavBarItemPadding()
+        setupRelationshipController()
+    }
+
+    private func setupRelationshipController() {
+        let chainableController = ResponderChainableController(
+            controller: self,
+            next: { [weak self] in
+                return self?.superNext
+            }
+        )
+
+        let relationshipController = RelationshipController()
+        relationshipController.responderChainable = chainableController
+        relationshipController.currentUser = self.currentUser
+        self.relationshipController = relationshipController
     }
 
     override func viewDidAppear(_ animated: Bool) {
@@ -48,7 +75,9 @@ class BaseElloViewController: UIViewController, HasAppController, ControllerThat
         UIApplication.shared.statusBarStyle = .lightContent
     }
 
-    func didSetCurrentUser() {}
+    func didSetCurrentUser() {
+        relationshipController?.currentUser = currentUser
+    }
 
     @IBAction
     func backTapped() {
--- a/Sources/Controllers/Relationship/RelationshipController.swift
+++ b/Sources/Controllers/Relationship/RelationshipController.swift
@@ -2,8 +2,8 @@
 ///  RelationshipController.swift
 //
 
-typealias RelationshipChangeClosure = (_ relationshipPriority: RelationshipPriority) -> Void
-typealias RelationshipChangeCompletion = (_ status: RelationshipRequestStatus, _ relationship: Relationship?, _ isFinalValue: Bool) -> Void
+typealias RelationshipChangeClosure = (_ relationshipPriority: RelationshipPriorityWrapper) -> Void
+typealias RelationshipChangeCompletion = (_ status: RelationshipRequestStatusWrapper, _ relationship: Relationship?, _ isFinalValue: Bool) -> Void
 
 class RelationshipRequestStatusWrapper: NSObject {
     let status: RelationshipRequestStatus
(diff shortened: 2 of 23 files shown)


Specs/Controllers/Relationship/RelationshipControlSpec.swift
```diff
@@ -8,19 +8,29 @@ import Nimble
...
             var relationshipController: RelationshipController!
+            let viewContainer = UIView()
+
             beforeEach {
                 subject = RelationshipControl()
-                presentingController = UIViewController()
+                presentingController = StreamableViewController()
+                presentingController.viewContainer = viewContainer
+                presentingController.viewDidLoad()
+                presentingController.view.addSubview(subject) // <-- super ghetto
```
steam: The responder chain is pretty magic. It took some creativity to get these specs to pass.

colinta: Apple really doesn't want us calling `viewDidLoad` directly like this, what about `showController` instead?

Sources/Controllers/Stream/StreamableViewController.swift
```diff
@@ -27,7 +28,12 @@ protocol InviteResponder: class {
     func sendInvite(person: LocalPerson, isOnboarding: Bool, completion: @escaping ElloEmptyCompletion)
 }
 
-class StreamableViewController: BaseElloViewController, PostTappedDelegate {
+struct ResponderChainableController {
+    let controller: UIViewController
```
colinta: I think this should be a `weak var`.  The `var responderChainable` is strongly held by objects which are strongly held by *this* controller.

Alternatively, `var responderChainable` could be `weak`, whichever is easiest.

Sources/Controllers/BaseElloViewController.swift
```diff
@@ -33,9 +33,36 @@ class BaseElloViewController: UIViewController, HasAppController, ControllerThat
...
+            controller: self,
+            next: { [weak self] in
+                return self?.superNext
+            }
+        )
+
+        let relationshipController = RelationshipController()
+        relationshipController.responderChainable = chainableController
+        relationshipController.currentUser = self.currentUser
+        self.relationshipController = relationshipController
```
colinta: As mentioned below, I think there might be a circular memory reference here:
```
StreamViewController → RelationshipController ↴
                  ↖︎ ResponderChainableController
```

Sources/Controllers/Relationship/RelationshipController.swift
```diff
@@ -2,43 +2,55 @@
...
 
-class RelationshipController {
+class RelationshipController: UIResponder {
     var currentUser: User?
-    weak var delegate: RelationshipControllerDelegate?
-    weak var presentingController: UIViewController?
+    var responderChainable: ResponderChainableController?
+
+    override var canBecomeFirstResponder: Bool {
+        return true
```
colinta: I'm curious to know if this `true` is required, or if it could return false; usually `true` here means that this object is capable of accepting keyboard, touch and motion events, not related to the responder chain resolution.

Sources/Controllers/Relationship/RelationshipController.swift
```diff
@@ -2,43 +2,55 @@
...
-// MARK: RelationshipController: RelationshipDelegate
-extension RelationshipController: RelationshipDelegate {
-    func relationshipTapped(_ userId: String, prev prevRelationshipPriority: RelationshipPriority, relationshipPriority: RelationshipPriority, complete: @escaping RelationshipChangeCompletion) {
-        Tracker.shared.relationshipButtonTapped(relationshipPriority, userId: userId)
-        if let shouldSubmit = delegate?.shouldSubmitRelationship(userId, relationshipPriority: relationshipPriority), !shouldSubmit {
+// MARK: RelationshipController: RelationshipResponder
+extension RelationshipController: RelationshipResponder {
+
+    func relationshipTapped(_ userId: String, prev prevRelationshipPriority: RelationshipPriorityWrapper, relationshipPriority: RelationshipPriorityWrapper, complete: @escaping RelationshipChangeCompletion) {
+        Tracker.shared.relationshipButtonTapped(relationshipPriority.priority, userId: userId)
```
colinta: There will be a conflict here w/ this bit of code: https://github.com/ello/ello-ios/pull/329/files#diff-3d1652560b2643c235e0cc41f75eea85

Easily resolved, just add the `guard` lines to the top of `func relationshipTapped`

Sources/Controllers/Stream/PostbarController.swift
```diff
@@ -409,12 +413,16 @@ class PostbarController: UIResponder, PostbarResponder {
 
         self.dataSource.insertUnsizedCellItems(items,
             withWidth: self.collectionView.frame.width,
-            startingIndexPath: commentsStartingIndexPath) { (indexPaths) in
+            startingIndexPath: commentsStartingIndexPath) {[weak self] (indexPaths) in
```
colinta: I noticed that in some places you use bracket-space-bracket, and in other places the two brackets are hugging

```
failure: { [weak self] in
..........^
startingIndexPath: commentsStartingIndexPath) {[weak self] (indexPaths) in
...............................................^
```

Sources/Controllers/Stream/PostbarController.swift
```diff
@@ -409,12 +413,16 @@ class PostbarController: UIResponder, PostbarResponder {
...
+                guard let `self` = self else { return }
                 self.collectionView.reloadData() // insertItemsAtIndexPaths(indexPaths)
                 cell.commentsControl.isEnabled = true
 
-                if indexPaths.count == 1 && jsonables.count == 0 {
-                    self.presentingController?.createCommentTapped(post.id)
+                if let controller = self.responderChainable?.controller,
+                    indexPaths.count == 1 && jsonables.count == 0
```
colinta: This could be *either*

```swift
if let controller = self.responderChainable?.controller,
    indexPaths.count == 1 && jsonables.count == 0  // as it is now
```
or
```swift
if let controller = self.responderChainable?.controller,
    indexPaths.count == 1, jsonables.count == 0  // representing && as multiple if conditions
```
Do we have a preference on this?

Sources/Controllers/Stream/StreamableViewController.swift
```diff
@@ -49,6 +54,8 @@ class StreamableViewController: BaseElloViewController, PostTappedDelegate {
 
     var scrollLogic: ElloScrollLogic!
 
+
+
```
colinta: That's a lot of newlines!

Specs/Controllers/Relationship/RelationshipControlSpec.swift
```diff
@@ -8,19 +8,29 @@ import Nimble
...
+                relationshipController = presentingController.relationshipController
                 showController(presentingController)
```
colinta: It's fine to call `showController` many times, it's just meant to setup the view of the controller (so `viewDidLoad` gets called)

colinta:
(sorry about the merge conflict!)

Sources/Controllers/Stream/PostbarController.swift
colinta: PS I noticed there will be a merge conflict here, too; just make sure `self.currentUser != nil` is added to this conditional, so that the "show omnibar if zero comments" feature doesn't activate in logged out.

Specs/Controllers/Relationship/RelationshipControlSpec.swift
steam: Turns out the `viewDidLoad()` call was extraneous. We call `showControlle()` a couple lines down. Its likely that the trickery of adding `RelationshipControl` to the view hierarchy fixed the issue. I removed the `viewDidLoad()` call entirely.

Sources/Controllers/Stream/StreamableViewController.swift
steam: Good catch, I've made the change. `ResponderChainableController` cannot be held weakly due to it being a struct value type. I changed the reference to `controller`, it is now weak and by necessity, optional.

Sources/Controllers/BaseElloViewController.swift
steam: 👍

Sources/Controllers/Stream/PostbarController.swift
steam: Interesting, I didn't know that multiple conditions could be added by separating them with commas. I think the comma approach seems fine. Although I will say, I miss the `where` keyword, imo it communicated intention much more clearly.

Sources/Controllers/Stream/PostbarController.swift
steam: Good eye. I'd noticed that when authoring the code and had meant to leave a space after the braces. I'll make them all consistent with a space between the curly and square brackets.

Sources/Controllers/Relationship/RelationshipController.swift
steam: In the diff you have the `guard` above the `Tracker...` call. Do we want to prevent tracking calls from going out for logged out users?

Sources/Controllers/Stream/StreamableViewController.swift
steam: 👀

Sources/Controllers/Relationship/RelationshipController.swift
steam: Looks like it is fine w/o the override. I'll remove it.

steam:
All changes made. Running specs now. You'll be happy to know that I merged `sd/feature/disable-logged-out-interaction` rather than rebasing @colinta. It was hard not to type `r-e-b-a-s-e`, but I did it! Also, thanks for the thorough review on this pr. Very helpful!

steam:
Executed 2623 tests, with 0 failures (0 unexpected) in 334.869 (337.185) seconds

steam:
Ok, if everything is good to go @colinta please approve it and we can merge this into `sd/feature/disable-logged-out-interaction`.

Sources/Controllers/Stream/PostbarController.swift
colinta: Yeah I miss `where`, and I'm not saying we *should* go the multiple commas route (I've been using it here and there), just wanted to point out that we have the option.  Now that you've seen both, do you have a preference?

Sources/Controllers/Relationship/RelationshipController.swift
colinta: We should have @caseydoran chime in – since the relationship control behavior is being bypassed I figured it would be better not to track that, but I could be wrong!

Sources/Controllers/Stream/PostbarController.swift
steam: I'm down to give the comma syntax a try. It still looks foreign to me but it wasn't long ago that `guard` felt unnatural. Now I love it. Sound good?

Sources/Controllers/Stream/ResponderChainableController.swift
```diff
@@ -0,0 +1,8 @@
+////
+///  ResponderChainableController.swift
```
colinta: It's a real boy!

colinta: (I mean it has its own file)

Sources/Controllers/Relationship/RelationshipController.swift
```diff
@@ -47,6 +47,11 @@ extension RelationshipController: RelationshipResponder {
         relationshipPriority: RelationshipPriorityWrapper,
         complete: @escaping RelationshipChangeCompletion)
     {
+		guard currentUser != nil else {
```
colinta: Xcode inserted two `\t` chars instead of 8 spaces. Because... yes, that's exactly what we wanted, I'm sure.  WTF XCODE!?

steam: Actually, filemerge did that. I thought I fixed it but I guess not. I'll fix it and merge this beast.

Sources/Controllers/Stream/ResponderChainableController.swift
steam: I was 💯 with you @colinta 👶

PR #337, opened 2017-02-09
Log tracking events depending on logged out event type

colinta:
@caseydoran thoughts on the names of these events?  They don't match webapp, because the app behavior isn't the same (we don't open a registration modal).


caseydoran:
Cool with me if @Justin-Holmes is cool with it. For me the key is being able to report on it.

PR #339, opened 2017-02-10
Show navigation bar on native screens launched from browser

steam:
Unfortunately this is only a partial fix for both of these bugs. 👇 

[Fixes #139398397](https://www.pivotaltracker.com/story/show/139398397) & [Fixes #139473309](https://www.pivotaltracker.com/story/show/139473309)

When launching a native screen from an in-app browser  the screen did not have a nav bar, it does now **_but_** the status bar does not hide correctly when scrolling. Further tweaking is necessary but I felt it was better to get a 1.22 build ready than delay it for this relatively minor bug.

Additionally, the screen can be launched in landscape mode if the browser is in landscape. Rotating to portrait will lock the screen to portrait and prevent further rotation. Similar to the first bug, this is a partial fix. Better than before but also requiring additional attention.

### DIFF
--- a/Sources/Controllers/Stream/StreamableViewController.swift
+++ b/Sources/Controllers/Stream/StreamableViewController.swift
@@ -55,6 +55,7 @@ class StreamableViewController: BaseElloViewController {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+        AppDelegate.restrictRotation = true
         showing = true
         willPresentStreamable(navigationBarsVisible())
     }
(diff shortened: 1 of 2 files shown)


Sources/Controllers/Stream/StreamableViewController.swift
```diff
@@ -55,6 +55,7 @@ class StreamableViewController: BaseElloViewController {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+        AppDelegate.restrictRotation = true
```
colinta: How does this behave on iPads?

colinta: Nevermind, looks like `AppDelegate` takes care of it.
```
        if UIDevice.current.userInterfaceIdiom == .phone {
            if AppDelegate.restrictRotation {
```

PR #343, opened 2017-02-13
Unify Following/Starred

steam:


Sources/Controllers/ElloTabBarController.swift
```diff
@@ -7,18 +7,18 @@ import SwiftyUserDefaults
...
-    static let DefaultTab = ElloTab.stream
+    static let DefaultTab: ElloTab = .following
 
     var narrationDefaultKey: String {
         let defaultPrefix = "ElloTabBarControllerDidShowNarration"
         switch self {
         case .discover:      return "\(defaultPrefix)Discover"
         case .notifications: return "\(defaultPrefix)Notifications"
-        case .stream:        return "\(defaultPrefix)Stream"
+        case .following:        return "\(defaultPrefix)Stream"
```
colinta: Whitespace ☝️

Sources/Controllers/Relationship/RelationshipControl.swift
```diff
@@ -219,21 +164,10 @@ class RelationshipControl: UIView {
...
-            starButton.frame = CGRect(x: frame.width - style.starButtonWidth, y: 0, width: style.starButtonWidth, height: Size.viewHeight)
-            starButtonWidth = style.starButtonWidth + style.starButtonMargin
-        }
-        else {
-            starButton.frame = .zero
-            starButtonWidth = 0
-        }
-
-        followingButton.frame = bounds.inset(top: 0, left: 0, bottom: 0, right: starButtonWidth)
+        followingButton.frame = bounds.inset(top: 0, left: 0, bottom: 0, right: 0)
```
colinta: This can just be `frame = bounds`

Sources/Utilities/InterfaceImage.swift
```diff
@@ -75,10 +75,6 @@ enum InterfaceImage: String {
     case bubbleBody = "bubble_body"
     case bubbleTail = "bubble_tail"
 
-    // Relationship
-    case whiteStar = "white_star"
-    case blackStar = "black_star"
```
colinta: Nice - we can probably remove the corresponding SVG files, yeah?

Sources/model/RelationshipPriority.swift
```diff
@@ -11,15 +11,14 @@ class RelationshipPriorityWrapper: NSObject {
 
 enum RelationshipPriority: String {
     case following = "friend"
-    case starred = "noise"
```
colinta: Since the API can still return `starred`, is it safe to remove it here?  Or is it possible to map `"noise"` to `.following`?

Sources/model/RelationshipPriority.swift
```diff
@@ -11,15 +11,14 @@ class RelationshipPriorityWrapper: NSObject {
...
     case inactive = "inactive"
     case none = "none"
     case null = "null"
     case me = "self"
 
-    static let all = [following, starred, block, mute, inactive, none, null, me]
+    static let all = [following, block, mute, inactive, none, null, me]
 
     init(stringValue: String) {
         self = RelationshipPriority(rawValue: stringValue) ?? .none
```
colinta: I wonder if it would be as simple as adding an `if/guard` here to `"noise" -> .following`

Sources/Controllers/Stream/FollowingViewController.swift
```diff
@@ -0,0 +1,139 @@
+////
+///  FollowingViewController.swift
+//
+
+import SwiftyUserDefaults
+
+class FollowingViewController: StreamableViewController {
+    override func trackerName() -> String? { return "Following" }
```
colinta: @Justin-Holmes for tracking purposes, should this return the same `Screen Stream, kind: following` that we are sending now?

Justin-Holmes: @colinta Yeah, that'd be awesome!

colinta:
Other than that this looks great!  Yay :red:

steam:
Its going to get more red. You just wait, @colinta!

Sources/Controllers/ElloTabBarController.swift
steam: 👌

Sources/Utilities/InterfaceImage.swift
steam: Yeah, odd, I thought I had. Good catch.

Sources/model/RelationshipPriority.swift
steam: Interesting, I assumed it wouldn't any longer. @alanpeabody, in what cases would we still get "starred"? Will the `relationship_priority` property on users remain "noise" if that is what it was originally created with? If so, I like you're idea of using a guard to remap the relationship to `.following` @colinta.

alanpeabody: yes, relationships are still being stored as friend and noise for the time being. If everything goes well and after iOS is "baked in" enough I could foresee normalizing them all.

steam: Cool, I implemented your approach @colinta. Nice and simple 👌 

```
init(stringValue: String) {
    if stringValue == "noise" {
        self = .following
    }
    else {
        self = RelationshipPriority(rawValue: stringValue) ?? .none
    }
}
```

steam:
Ready for final review @colinta 🤘

PR #347, opened 2017-02-21
Select comment icon if content is present

steam:
[Fixes #138356549](https://www.pivotaltracker.com/story/show/138356549)


Sources/Controllers/Stream/CellDequeing/StreamFooterCellPresenter.swift
```diff
@@ -73,10 +73,18 @@ struct StreamFooterCellPresenter {
...
-        streamKind: StreamKind )
+        streamKind: StreamKind,
+        currentUser: User?)
     {
-        if streamKind.isDetail {
+        if streamKind.isDetail && currentUser == nil {
+            if let count = post.commentsCount, count > 0 {
+                cell.commentsOpened = true
+                cell.commentsControl.isSelected = true
+            }
```
colinta: Is `cell.commentsOpened` reset in `prepareForReuse`?  Otherwise it will be in a "reused" state here.

steam:
Good catch @colinta - changed it so the correct value is always assigned.

PR #353, opened 2017-02-27
Updated Navigation

steam:
[Finishes #139394909](https://www.pivotaltracker.com/story/show/139394909)

screen shot 2017-02-27 at 2 30 35 pm


Sources/Controllers/Profile/ProfileViewController.swift
```diff
@@ -335,11 +348,15 @@ extension ProfileViewController {
...
-        guard let cachedImage = cachedImage(.coverImage) else {
-            return
+        if
+            let cachedAvatar = cachedImage(.avatar),
+            let item = tabBarItem as? AvatarBarItem
+        {
+            item.setUserAvatar(cachedAvatar)
+        }
+        else if let cachedCoverImage = cachedImage(.coverImage) {
+            screen.coverImage = cachedCoverImage
```
rynbyjn: It seems odd that both of these conditions don't run since the cover and avatar can co-exist, but maybe I'm missing something..

Specs/Utilities/InterfaceSpec.swift
```diff
@@ -93,7 +93,7 @@ class InterfaceSpec: QuickSpec {
                         .bolt,
                         .omni,
```
rynbyjn: does `omni` need to get updated to `create_post` since it was changed in the `InterfaceImage` file?

Sources/Controllers/Profile/ProfileViewController.swift
steam: In practice this function will never get called once with both changing but you're right, there is no reason to make it exclusive. I'll update it.

Specs/Utilities/InterfaceSpec.swift
steam: I didn't think it needed to be changed here and at every call site. The `InterfaceImage` file references a different asset but keeps the name used throughout the app the same.

PR #360, opened 2017-03-06
Tracks post detail page views when viewing a cached post

colinta:
Adds `sendPostView` to PostDetailGenerator and `postView` endpoint to ElloAPI

`sendPostView` is only called if the post was loaded from memory, and not during a reload
postView is just one possible implementation of the `post_views` endpoint, and it
hard codes the values of `kind`.  In the future we could change it to support different
`post_views.kind` values.

Finishes [#140890503](https://www.pivotaltracker.com/story/show/140890503)

### DIFF
--- a/Sources/Controllers/Stream/Detail/PostDetailGenerator.swift
+++ b/Sources/Controllers/Stream/Detail/PostDetailGenerator.swift
@@ -80,7 +80,7 @@ private extension PostDetailGenerator {
     func sendPostView(_ doneOperation: AsyncOperation, reload: Bool = false) {
         guard let post = post, doneOperation.isFinished && !reload else { return }
 
-        PostService().sendPostView(id: post.id, token: post.token, email: currentUser?.profile?.email)
+        PostService().sendPostDetailView(id: post.id, email: currentUser?.profile?.email)
     }
 
     func loadPost(_ doneOperation: AsyncOperation, reload: Bool = false) {
--- a/Sources/Controllers/Stream/Detail/PostDetailGenerator.swift
+++ b/Sources/Controllers/Stream/Detail/PostDetailGenerator.swift
@@ -80,7 +80,7 @@ private extension PostDetailGenerator {
     func sendPostView(_ doneOperation: AsyncOperation, reload: Bool = false) {
         guard let post = post, doneOperation.isFinished && !reload else { return }
 
-        PostService().sendPostDetailView(id: post.id, email: currentUser?.profile?.email)
+        PostService().sendPostDetailView(id: post.id, userId: currentUser?.id)
     }
 
     func loadPost(_ doneOperation: AsyncOperation, reload: Bool = false) {
--- a/Sources/Controllers/Stream/Detail/PostDetailGenerator.swift
+++ b/Sources/Controllers/Stream/Detail/PostDetailGenerator.swift
@@ -43,6 +43,7 @@ final class PostDetailGenerator: StreamGenerator {
             setPlaceHolders()
         }
         setInitialPost(doneOperation)
+        sendPostView(doneOperation, reload: reload)
         loadPost(doneOperation, reload: reload)
         displayCommentBar(doneOperation)
         loadPostComments(doneOperation)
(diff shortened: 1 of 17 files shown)


steam:
If I remember correctly @jayzes had mentioned an endpoint for sending an array of post ids to be tracked in a single call. Is that a separate endpoint than this one?

colinta:
It's the same one, they are sent as a comma separated list.  From the specs:

```ruby
let(:posts) { "#{post1.token},#{post2.token},#{post3.token}" }
```

colinta:
```Executed 2556 tests, with 0 failures```

**Comment by ellodanger** [comment]
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

**Comment by colinta** [comment]
@steam I added the "track all posts on view appeared" code to this PR, so maybe a re-review of that?

File: Sources/Controllers/Stream/StreamableViewController.swift [hunk 1]
```diff
@@ -80,6 +80,20 @@ class StreamableViewController: BaseElloViewController {
...
 
+    func trackerStreamInfo() -> (String, String?)? {
+        return nil
+    }
+
+    override func trackScreenAppeared() {
+        super.trackScreenAppeared()
+
+        guard let (streamKind, streamId) = trackerStreamInfo() else { return }
+        let posts = streamViewController.dataSource.visibleCellItems.flatMap { streamCellItem in
```
steam: This is likely to bring back duplicates of the same post due to the one to many nature of the jsonables to stream cell items. I think we need to filter the results by group or something similar.

colinta: Good catch, fixing.

colinta:
I'll double check that we don't double track, but reloading doesn't (shouldn't?) call `viewDidAppear`, since the view is already there, we're just clearing the items and showing the loader.  And post views get tracked on the API side, so the order of events should be:

1. Open the app and view "Following"
1. `viewDidAppear` gets called, but no posts are loaded yet, so `PostService().sendPostViews` has no effect
1. Posts are requested from the API, which increases their post view count.
1. Scrolling on following happens, as usual.
1. Tap on "Discover"
1. `viewDidAppear` gets called, but again no posts are loaded
1. Posts are requested, post view count increases
1. Tap on "Following" again
1. `viewDidAppear` gets called and posts *are* available, so the post view count is increased.

I just realized, though, that we are not sending an `ExperienceUpdate` in this case; should we increase the view counts when we send `PostService().sendPostViews`?  I'll hit up @caseydoran on Slack about this, too.

Sources/Controllers/Stream/Detail/PostDetailGenerator.swift
```diff
@@ -76,6 +77,12 @@ private extension PostDetailGenerator {
         }
     }
 
+    func sendPostView(_ doneOperation: AsyncOperation, reload: Bool = false) {
+        guard let post = post, doneOperation.isFinished && !reload else { return }
+
+        PostService().sendPostDetailView(id: post.id, email: currentUser?.profile?.email)
```
jayzes: Did we look into modifying the endpoint to accept either the user id or e-mail address? I'd love to not send the e-mail address around unnecessarily if we can avoid it. /cc @Justin-Holmes

colinta: Oooh good point.

Justin-Holmes: @jayzes @colinta Sounds good, I just made a note to tweak the endpoint to accept `user_id` and `post_ids`

jayzes: 🤘 thanks @Justin-Holmes !

colinta:
@steam @jayzes Updates: sending `user_id` to endpoint instead of `email`, and only sending unique post ids (`Set<String>` instead of `Array<String>`).

jayzes:
👍

PR #361, opened 2017-03-09
Make staging and prod keys/secrets available during startup

colinta:
Instead of `rake generate:staging_keys or prod_keys`, you just run `rake
generate:keys` and it will store staging *and* production info.

Switch to staging and back to production via `DebugController`.  If you change
API servers the app logs you out and crashes itself, because some things
(Segment) are assigned a secret key only once, and rather than fix all those
places it made sense to just restart everything from fresh.

This should prevent:

- releasing a build that points to staging
- releasing a build that uses the wrong segment key

and enable:

- testing something on staging without recompiling the app
- test staging from a device running the production app

[Finishes #140890503](https://www.pivotaltracker.com/story/show/140890503)


colinta:
```Executed 2553 tests, with 0 failures```

PR #363, opened 2017-03-10
Use mp4 videos instead of gifs if available

steam:
The API now provides `.mp4` files for some gifs with the goal of reducing file size. This PR implements `.mp4` video in place of gifs using `AVPlayer`. 

Additionally:
* updated PINRemoteImage to newer version
* fixes an issue where image cells were reporting a height mismatch do to floating point math imprecision
* refactored some commonly repeated code out of presenters/cells into `Asset`

`Executed 2570 tests, with 0 failures (0 unexpected)`


colinta:
I just noticed a couple places where the code is ```guard result.hasImage == true else```, is `hasImage` a `Bool?`, or could these be `guard result.hasImage else?

Sources/Controllers/Stream/Cells/StreamImageCell.swift
```diff
@@ -301,3 +357,20 @@ class StreamImageCell: StreamRegionableCell {
...
 }
+
+extension StreamImageCell: DismissableCell {
+
+    func didEndDisplay() {
+        player.pause()
+        foregroundObserver?.removeObserver()
+        if let observer = videoObserver {
+            NotificationCenter.default.removeObserver(observer)
+        }
```
colinta: I wonder if these two observers will not be removed in some cases, e.g. pressing the back button.  The `weak self` guards against strange behavior, so I don't think it would be a source of bugs, but we could potentially have a lot of observers that never get released.

Sources/Model/Regions/Asset.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 import SwiftyJSON
 
 
-let AssetVersion = 1
+let AssetVersion = 2
```
colinta: Since the new property is optional you don't need to version it - plus, there is no version check in the NSCoding `init(coder:)` method, which is where that version number is used to guard against trying to get data that was never encoded.

steam: Good call.

steam:
Changing the `guard result.hasImage = true else ...` to `guard result.hasImage else ...`.

PR #364, opened 2017-03-10
Fix failing specs

steam:
marks slow spec pending

I left it pending @colinta, it is definitely worth looking into some day. Its just time to move on at the moment.


steam:
Fixed in #367

PR #365, opened 2017-03-10
Loads xhdpi instead of hdpi on iPad

steam:
Single column streams on screens 768pts or wider now load the `xhdpi` asset url instead of the `hdpi` one. This gets us most of the way to full-width images on iPad. Further explorations in scaling are needed. They've been captured in the roadmap card https://trello.com/c/8SU0XRnV/263-improve-the-ios-app-experience-on-ipad.

[Finishes #140426449](https://www.pivotaltracker.com/story/show/140426449)


steam:
Executed 2576 tests, with 0 failures (0 unexpected) in 242.655 (244.169) seconds

PR #386, opened 2017-03-31
Refactor interface image assignment

colinta:
make assignment shorter, support `InterfaceImage.Style`

closer to how `UIButton` assigns images.

[Finishes #142860715](https://www.pivotaltracker.com/story/show/142860715)

### DIFF
--- a/Sources/Extensions/UIImageViewExtensions.swift
+++ b/Sources/Extensions/UIImageViewExtensions.swift
@@ -3,9 +3,13 @@
 //
 
 extension UIImageView {
+    var interfaceImage: InterfaceImage? {
+        get { return nil }
+        set { setInterfaceImage(newValue!, style: .normal) }
+    }
 
-    func setImage(_ interfaceImage: InterfaceImage, degree: Double) {
-        self.image = interfaceImage.normalImage
+    func setInterfaceImage(_ interfaceImage: InterfaceImage, style: InterfaceImage.Style, degree: Double = 0) {
+        self.image = interfaceImage.image(style)
         if degree != 0 {
             let radians = (degree * Double.pi) / 180.0
             self.transform = CGAffineTransform(rotationAngle: CGFloat(radians))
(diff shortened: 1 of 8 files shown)


ellodanger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/Extensions/UIImageViewExtensions.swift
```diff
@@ -3,9 +3,13 @@
 //
 
 extension UIImageView {
+    var interfaceImage: InterfaceImage? {
+        get { return nil }
```
steam: I understand why this is `nil` but I'm concerned it might confuse us in the future. I don't have a good solve for it but am curious if you thought about it.

colinta: I had a few ideas at the time of writing:
```swift
// store the value, even though it's never used, and would be out of
// sync if the image is set without using this helper
var interfaceImage: InterfaceImage? {
  didSet {
    if let interfaceImage = interfaceImage {
      setInterfaceImage(newValue, style: .normal)
    }
  }
}

// error on "getting", at a _small_ risk that code would be written
// that gets the interfaceImage and isn't tested before being released
var interfaceImage: InterfaceImage {
  get { fatalError("'get' is not supported on UIImageView.interfaceImage") }
  set {
      setInterfaceImage(newValue, style: .normal)
  }
}
```

I went with the `return nil` because it seemed safer than a fatalError, and would catch a programmer's attention if they were trying to use this property in an "unsupported" way.

colinta: An even better approach might be to use a `StyledImage` that has this property, which is an even better flag to devs that we are doing something special - and we could set `interfaceImage` to `nil` if `var image: UIImage?` is set directly.

At the end of the day, it's *also* strange to have `interfaceImage` *and* `setInterfaceImage(style:)`.

steam: Ideally we'd be able to return `interfaceImage` if one had been set. The lack of stored properties in extensions prevents it. Overall I think this refactor is an improvement. I say ship it if you're happy with it.

PR #387, opened 2017-04-03
Remove mp4 video, use gifs instead

steam:
We had hoped that using mp4s on iOS, similar to what we do on web, would have improved the user experience by speeding up animation downloads. Unfortunately, mp4 video brought a host of other issues. Namely, cell flicker when reloading cells along with intermittent decoding issues when several animations were onscreen at the same time.

I feel like I’ve completed a pretty exhaustive exploration of downloading mp4s for animations on iOS. I spiked on an approach where I cached a programmatically generated thumbnail of the mp4 in addition to the AVURLAsset for the mp4. The result was a _somewhat_ improved experience in the form of _a bit_ less flicker. The issue of AVFoundation reporting decoding errors did not go away unfortunately. Our best guess for the decoding errors is the presence of too many AVPlayer instances, one for each mp4 on screen which can be quite a few in grid mode. I know of no way to get around this issue other than having less AVPlayers which means less animations playing at once.

Another exploration involved downloading mp4s and transcoding them into gifs on the device. Transcoding a single, short mp4 to a gif can be relatively quick (<2 seconds). Transcoding longer mp4s and queueing up lots of videos to transcode results in extremely long transcoding times (>45 seconds in some cases). This is with a robust cache in place as well. Viewing the gif category in grid mode requires queueing up dozens of mp4s for transcoding.

I should mention, for completeness, that I spent awhile chatting with the lead iOS dev at Giphy. They came to the same decision, not to use mp4s when you need to show a lot of animations at the same time on the same screen. They use a combination of webp, gif and mp4 depending on the scenario. For example, on post details with a small number of gifs playing we could load an mp4. If we’ve already loaded the gif however it doesn’t make much sense to download _another_ file. For large gifs that we don't play in non-post detail streams we could explore loading a much smaller mp4 on the post detail. That should work well.


ellodanger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Nice, more deletions than insertions :red_circle:</td>
    </tr>
  </tbody>
</table>

</p>

PR #388, opened 2017-04-03
Adds `More Posts` UI to Following

colinta:
- added `NewPostsButton` class, updated `arrow.svg` to match comps (pulled in from Sketch, basically).
- shows on `newFollowingContent` notification
- hides on scroll-to-top
- tapping the button reloads `FollowingViewController`

Unrelated refactors:

- updated controllers and views to use `interfaceImage` setter (dunno why this wasn't in #386, I thought I had pulled these changes into that PR)
- `NewContentNotifications` use `()` instead of `Void?` as their value (removed `_ in` from listener blocks)
- fixed the notification listeners in `FollowingViewController`, they were not all being turned on and off.
- refactored `lastViewedCreatedAtKey` to be optional, and unique on a per-discover-endpoint basis (rather than based on `cacheKey`)

[Finishes #141638127](https://www.pivotaltracker.com/story/show/141638127)

### DIFF
--- a/Sources/Controllers/Stream/FollowingViewController.swift
+++ b/Sources/Controllers/Stream/FollowingViewController.swift
@@ -34,6 +34,10 @@ class FollowingViewController: StreamableViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
+    deinit {
+        removeNotificationObservers()
+    }
+
     override func viewDidLoad() {
         super.viewDidLoad()
         view.backgroundColor = .white
@@ -44,6 +48,8 @@ class FollowingViewController: StreamableViewController {
         streamViewController.streamKind = .following
         ElloHUD.showLoadingHudInView(streamViewController.view)
         streamViewController.loadInitialPage()
+
+        addNotificationObservers()
     }
 
     override func viewDidAppear(_ animated: Bool) {
@@ -58,7 +64,7 @@ class FollowingViewController: StreamableViewController {
 
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
-        removeNotificationObservers()
+        removeTemporaryNotificationObservers()
     }
 
     override func viewForStream() -> UIView {
--- a/Sources/Controllers/Stream/FollowingViewController.swift
+++ b/Sources/Controllers/Stream/FollowingViewController.swift
@@ -2,6 +2,7 @@
 ///  FollowingViewController.swift
 //
 
+import SnapKit
 import SwiftyUserDefaults
 
 
@@ -15,6 +16,7 @@ class FollowingViewController: StreamableViewController {
     }
 
     var navigationBar: ElloNavigationBar!
+    let newPostsButton = NewPostsButton()
     fileprivate var loggedPromptEventForThisSession = false
     fileprivate var reloadFollowingContentObserver: NotificationObserver?
     fileprivate var appBackgroundObserver: NotificationObserver?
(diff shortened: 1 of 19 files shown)


Sources/Controllers/Stream/FollowingViewController.swift
```diff
@@ -97,6 +113,27 @@ class FollowingViewController: StreamableViewController {
         self.present(drawer, animated: true, completion: nil)
     }
 
+    override func streamViewDidScroll(scrollView: UIScrollView) {
+        super.streamViewDidScroll(scrollView: scrollView)
+
+        if scrollView.contentOffset.y <= 0 {
+            animate {
+                self.newPostsButton.alpha = 0
```
steam: Will the button still be tappable if we don't disable or hide it entirely?

Sources/Networking/NewContentService.swift
```diff
@@ -88,34 +89,34 @@ private extension NewContentService {
...
 
                  done()
              },
              failure: { _ in done() })
     }
 
-    func checkForNewStreamContent(_ done: @escaping BasicBlock = {}) {
-        let storedKey = StreamKind.following.lastViewedCreatedAtKey
+    func checkForNewFollowingContent(_ done: @escaping BasicBlock = {}) {
+        let storedKey = StreamKind.following.lastViewedCreatedAtKey!
```
steam: I trust that this won't crash in this particular instance but it seems unsafe to have this optional and requiring a force unwrap.

ellodanger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR</td>
    </tr>
  </tbody>
</table>

</p>

Sources/Networking/NewContentService.swift
colinta: I had `guard` here at first, but then I thought: (a) the force-unwrap is covered in specs, and we *should* truest that the key is available.  (b) It would be a gotcha!-style-bug if the code changed and the guard dropped out of this method because the key was `nil`.  Ideally we wouldn't need an optional, but I like it as an optional because there really *are* endpoints that this key doesn't apply to, like post detail and profiles.

Sources/Controllers/Stream/FollowingViewController.swift
colinta: There's a threshold, like `0.01` or so, where a control is no longer interactive (I don't recall, but I *think* it'll return `true` for `isHidden`, too).  I don't remember the number, but IIRC this prevents tapping.  Double checking now.

colinta: Verified it works as expected, but it doesn't affect `isHidden` or `isUserInteractionEnabled` or `isEnabled`. The `hitTest(with:)` method returns `nil` when `alpha <= 0.01`.

steam: Good to know, I'd have put in the extra disabling code. No need!

Sources/Networking/NewContentService.swift
steam: Fair enough. My force unwrap hackles are always raised after the slew of crashers we had earlier on.

colinta: For sure, and they should always be justifiable (and, unless it's not possible, spec'd).

PR #389, opened 2017-04-03
Adds build number and version to all tracking events

colinta:
and show the version in the drawer from the sim (unless running specs) and in the debug controller

[Finishes #141359233](https://www.pivotaltracker.com/story/show/141359233)


jayzes:
Do we need the extra params in addition to what the [Segment SDK already provides](https://segment.com/docs/spec/common/#context-fields-automatically-collected)? I think it might get these two fields already.

colinta:
Oh snap, that's handy!  Thanks for catching that.

PR #392, opened 2017-04-06
Fix tracking on related posts

colinta:
In order to accomplish this I had to "re-insert" the `StreamViewController` into the responder.  Easily done by adding a `properTarget(forAction:,withSender:)` to `PostbarController`.  This method starts the responder chain search at `StreamViewController` instead of `PostbarController`.

The `PostbarController` had a bit of schizophrenia going on whether to use `cell` or `indexPath`, so I refactored the code to prefer `cell`.  This also simplified the `StreamHeader/FooterCell` code.

[Finishes #142723269](https://www.pivotaltracker.com/story/show/142723269)

### DIFF
--- a/Sources/Controllers/Stream/Cells/StreamCreateCommentCell.swift
+++ b/Sources/Controllers/Stream/Cells/StreamCreateCommentCell.swift
@@ -180,17 +180,13 @@ class StreamCreateCommentCell: UICollectionViewCell {
     }
 
     func replyAllTapped() {
-        guard let indexPath = indexPath else { return }
-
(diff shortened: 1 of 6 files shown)


colinta:
I liked the refactor until I had to fix the specs - the `indexPath(for: cell)` call doesn't work so well when you can't control the `UICollectionView.visibleCells`.  Instead, I added a `fakeIndexPath`, which is settable from the specs.

steam:
Ha, maybe thats why we went with `indexPath` in the first place.

PR #397, opened 2017-04-10
Use smaller images in more places

colinta:
Decrease the memory pressure by using smaller avatars and category images.

The other code refactors were due to just poking around, trying different memory changes.

[Finishes #141965649](https://www.pivotaltracker.com/story/show/141965649)
[Finishes #143336577](https://www.pivotaltracker.com/story/show/143336577)


colinta:
Yup, crashed every time.  Must be an even older ipad.

PR #400, opened 2017-04-13
Colors announcement cell red when staff

colinta:
Makes staff-preview announcements red.

[Finishes #143349211](https://www.pivotaltracker.com/story/show/143349211)


Sources/Model/Announcement.swift
```diff
@@ -63,13 +75,15 @@ final class Announcement: JSONAble, Groupable {
     override class func fromJSON(_ data: [String: AnyObject]) -> JSONAble {
         let json = JSON(data)
         let id = json["id"].stringValue
+        let isStaffPreview = json["is_staff_preview"].boolValue
```
colinta: @alanpeabody Another API sanity check `is_staff_preview` on `announcement`.

alanpeabody: Yup, It is on all environments now too if you want to test it out somewhere.

PR #413, opened 2017-04-25
double tap love

colinta:
[Finishes #144123155](https://www.pivotaltracker.com/story/show/144123155)

### DIFF
--- a/Sources/Controllers/Stream/StreamViewController.swift
+++ b/Sources/Controllers/Stream/StreamViewController.swift
@@ -892,6 +892,7 @@ extension StreamViewController: StreamEditingResponder {
         }
 
         if !post.loved {
+            Tracker.shared.doubleTapLove(post)
             let footerCell = collectionView.cellForItem(at: footerPath) as? StreamFooterCell
             postbarController?.toggleLove(footerCell, post: post)
         }
--- a/Sources/Controllers/Stream/StreamViewController.swift
+++ b/Sources/Controllers/Stream/StreamViewController.swift
@@ -892,9 +892,8 @@ extension StreamViewController: StreamEditingResponder {
         }
 
         if !post.loved {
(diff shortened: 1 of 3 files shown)


Sources/Controllers/Stream/StreamViewController.swift
```diff
@@ -892,6 +892,7 @@ extension StreamViewController: StreamEditingResponder {
         }
 
         if !post.loved {
+            Tracker.shared.doubleTapLove(post)
```
jayzes: Does this track in addition to "post loved" or instead of it? Would it be possible to use a single event name but add an attribute indicating that t came from double-tap? I think that might make the dashboarding more intuitive. /cc @Justin-Holmes @caseydoran

colinta: this is an additional tracking; so we'd have to compare the two events to get how many people double-tapped.

colinta: I added the `via: "button"|"double tap"` property.  I left the `double tap love` in place, too, in case it's handy.

caseydoran: I like the attribute option, feels cleaner

Sources/Utilities/Tracker.swift
```diff
@@ -480,8 +480,12 @@ extension Tracker {
         track("Post shared", properties: ["post_id": post.id])
     }
 
-    func postLoved(_ post: Post) {
-        track("Post loved", properties: ["post_id": post.id])
+    func doubleTapLove(_ post: Post) {
```
steam: It looks like `doubleTapLove` was added but never called.

colinta: Oh yeah I said I was gonna keep using it, then I deleted it.  I'll remove it, casey says we don't need it.

PR #414, opened 2017-04-25
Adds new relic, assumes .env key 'NEW_RELIC_KEY'

colinta:
[Finishes #140890039](https://www.pivotaltracker.com/story/show/140890039)


colinta:
Oh yeah, that should be straightforward.

PR #416, opened 2017-04-27
Conversion and identify changes

colinta:
- track login and join button taps
- track screen views (as events) when logged out
- add `is_featured` and `is_nabaroo` to identify properties

[Finishes #143615505](https://www.pivotaltracker.com/story/show/143615505)
[Finishes #144393253](https://www.pivotaltracker.com/story/show/144393253)
[Finishes #144370967](https://www.pivotaltracker.com/story/show/144370967)
[Finishes #144065163](https://www.pivotaltracker.com/story/show/144065163)

### DIFF
--- a/Sources/Utilities/Tracker.swift
+++ b/Sources/Utilities/Tracker.swift
@@ -172,7 +172,7 @@ extension Tracker {
     }
 
     func loginButtonTapped() {
-        track("login link tapped")
+        track("login button tapped")
     }
 
     func loginValid() {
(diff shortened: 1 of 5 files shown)


Sources/Utilities/Tracker.swift
```diff
@@ -143,9 +130,9 @@ extension Tracker {
         track("tapped about")
     }
 
-    func tappedTsAndCs() {
-        track("tapped terms and conditions")
-    }
+    // func tappedTsAndCs() {
```
steam: Should this be removed entirely?

colinta: When I left it, I was thinking that if we _readded_ this tracking event, we should name it the same as what it was before.  On second thought, why not _wire it back up!?_  So I did.

PR #425, opened 2017-05-12
Adds "All" category

colinta:
Tapping 'All' shows a back button, which returns you to the previous category.  It hides the category card list (because it's redundant) and the share and grid toggle buttons.

[Finishes #145231965](https://www.pivotaltracker.com/story/show/145231965)


colinta:
```
Test Suite 'All tests' passed at 2017-05-16 11:03:36.399.
	 Executed 3841 tests, with 0 failures (0 unexpected) in 429.347 (430.892) seconds
```

PR #427, opened 2017-05-16
Add 'is_experimental' to Segment identify

colinta:
[Finishes #145311155](https://www.pivotaltracker.com/story/show/145311155)


colinta:
@jayzes I looked in the JWT and didn't see `experimental` in there, but we have it on `User`, which we load before calling `Tracker.shared.identify` anyway.

PR #431, opened 2017-05-25
Show and Hide the lovers drawer after love/unlove

colinta:
If the current user is the only lover, show the drawer.  If no users love the post, hide the drawer.

[Finishes #145882805](https://www.pivotaltracker.com/story/show/145882805)


colinta:
I did my best to reuse code by putting `StreamCellItem` generators into static methods, but the logic of showing/hiding the drawer, and inserting/deleting the *"social padding"* cell is unfortunately duplicated.

PR #432, opened 2017-05-30
Switch to PromiseKit

colinta:
I love the syntax and features, and the services look so much better now!

One gotcha: PromiseKit's promises usually complete on the "next tick", even when the value is available immediately, this had terrible repurcussions for our specs.  The workaround is to use PromiseKit's `zalgo` dispatch queue, which calls its `then` blocks synchronously.

[Finishes #146322753](https://www.pivotaltracker.com/story/show/146322753)

### DIFF
--- a/Sources/Controllers/Editorials/EditorialsGenerator.swift
+++ b/Sources/Controllers/Editorials/EditorialsGenerator.swift
@@ -48,7 +48,7 @@ private extension EditorialsGenerator {
 
         let receivedEditorials = afterAll()
         StreamService().loadStream(streamKind: streamKind)
-            .onSuccess { [weak self] response in
+            .thenFinally { [weak self] response in
                 guard
                     let `self` = self,
                     case let .jsonables(jsonables, responseConfig) = response,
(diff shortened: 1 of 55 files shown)


colinta:
@steam the block behavior of PromiseKit is deliberate, and for most of our app it's fine.  It's meant to avoid inconsistencies like this:

```swift
var x = 0
foo.then {
    // is `foo` resolved?  if it is, this block will be
    // called immediately, otherwise it will be called later
    print(x)  // => 0 or 1
}
x++
```

Using zalgo allows this undefined behavior, and so our specs pass.  In general we don't need to use it, calling the blocks on the next tick is fine.

Sources/Controllers/Editorials/EditorialsGenerator.swift
```diff
@@ -49,7 +49,7 @@ private extension EditorialsGenerator {
         let receivedEditorials = afterAll()
         ElloProvider.oneTimeProvider = ElloProvider.StubbingProvider()
         StreamService().loadStream(streamKind: streamKind)
-            .onSuccess { [weak self] response in
+            .thenFinally { [weak self] response in
```
colinta: an alternative to this `thenFinally` helper method is

```swift
.then { response -> Void in
}

// then finally just makes -> Void explicit in the function signature, no need to declare it here
.thenFinally { response in
}
```

PR #437, opened 2017-06-15
Refactor Services to use Promises

colinta:
[Finishes #147263097](https://www.pivotaltracker.com/story/show/147263097)


Sources/Controllers/Relationship/RelationshipController.swift
```diff
@@ -79,36 +79,28 @@ extension RelationshipController: RelationshipResponder {
...
-                    let responder = self.target(forAction: #selector(RelationshipControllerResponder.relationshipChanged(_:status:relationship:)), withSender: self) as? RelationshipControllerResponder
-                    responder?.relationshipChanged(userId, status: RelationshipRequestStatusWrapper(status: .success), relationship: nil)
+        let responder: RelationshipControllerResponder? = self.findResponder()
+        let (optimisticRelationship, promise) = RelationshipService().updateRelationship(currentUserId: currentUserId, userId: userId, relationshipPriority: newRelationshipPriority.priority)
+        if let relationship = optimisticRelationship {
+            complete(RelationshipRequestStatusWrapper(status: .success), relationship, false)
+            responder?.relationshipChanged(userId, status: RelationshipRequestStatusWrapper(status: .success), relationship: relationship)
+        }
 
+        promise.thenFinally { (relationship) in
```
rynbyjn: Could there be a possibility that the promise returned from `updateRelationship` would resolve before the `thenFinally` is tacked on?

Sources/Controllers/Settings/DeleteAccountConfirmationViewController.swift
```diff
@@ -75,15 +75,16 @@ class DeleteAccountConfirmationViewController: BaseElloViewController {
...
-                postNotification(AuthenticationNotifications.userLoggedOut, value: ())
+        ProfileService().deleteAccount()
+            .thenFinally {
+                Tracker.shared.userDeletedAccount()
+                self.dismiss(animated: true) {
+                    postNotification(AuthenticationNotifications.userLoggedOut, value: ())
+                }
+            }
+            .always { _ in
+                ElloHUD.hideLoadingHud()
```
rynbyjn: Nice!

Sources/Networking/ContentFlaggingService.swift
```diff
@@ -4,17 +4,13 @@
...
 
-    init(){}
-
-    func flagContent(_ endpoint: ElloAPI, success: @escaping ContentFlaggingSuccessCompletion, failure: @escaping ElloFailureCompletion) {
-        ElloProvider.shared.elloRequest(endpoint,
-            success: { data in
-                success()
-        }, failure: failure)
+    func flagContent(_ endpoint: ElloAPI) -> Promise<Void> {
+        return ElloProvider.shared.request(endpoint).asVoid()
```
rynbyjn: I'm guessing that `asVoid` is because this is a fire and forget type of request?

Sources/Networking/NewContentService.swift
```diff
@@ -93,55 +102,50 @@ private extension NewContentService {
...
-            success: { (_, responseConfig) in
-                if let statusCode = responseConfig.statusCode, statusCode == 204 {
-                    postNotification(NewContentNotifications.newNotifications, value: ())
-                }
+        return ElloProvider.shared.request(.notificationsNewContent(createdAt: storedDate))
+            .thenFinally { response in
+                guard
+                    let statusCode = response.1.statusCode,
+                    statusCode == 204
+                    else { return }
```
rynbyjn: I dig the things you can do with `guard` like even just checking a condition.

Sources/Networking/S3UploadingService.swift
```diff
@@ -2,39 +2,41 @@
...
         }
         else {
-            upload(image.image, success: success, failure: failure)
+            return upload(image.image)
         }
     }
 
-    func upload(_ image: UIImage, success: @escaping S3UploadSuccessCompletion, failure: @escaping ElloFailureCompletion) {
+    func upload(_ image: UIImage) -> Promise<URL?> {
+        let (promise, fulfill, reject) = Promise<URL?>.pending()
```
rynbyjn: Why they have to use `fulfill` instead of `resolve`?

Sources/Controllers/Relationship/RelationshipController.swift
colinta: Actually, *no!*  PromiseKit has a condition that resolutions and rejections all happen on the *next* tick, to avoid situations like that.  Actually, getting the promises to run "synchronously" in the specs was an early pain point (see "zalgo" for how it was resolved).

Sources/Controllers/Settings/DeleteAccountConfirmationViewController.swift
colinta: Yeah, love the `always` option.

Sources/Networking/ContentFlaggingService.swift
colinta: Exactly - it says "still a promise, but return type isn't relevant"

Sources/Networking/S3UploadingService.swift
colinta: Yeah PromiseKit calls it `fulfill` internally, and I stuck w/ it.  Sorry!

colinta:
@rynbyjn I might get rid of `thenFinally`, I'm already regretting it.  It's just a way to get around swift's compiler - it declares the closure inside as returning `Void` (which also means it transforms the `Promise<T>` into `Promise<Void>`), but most importantly it means you don't have to return anything from the closure.

Another way to accomplish the same thing:

```swift
promise.then { _ -> Void in
   ...
}
promise.thenFinally { _ in
   ...
}
```

PR #499, opened 2017-10-06
Adds timer and countdown to Artist Invites

colinta:
[Finishes #151694644](https://www.pivotaltracker.com/story/show/151694644)


Sources/Controllers/ArtistInvites/ArtistInviteHeaderCell.swift
```diff
@@ -107,28 +107,36 @@ class ArtistInviteHeaderCell: CollectionViewCell, ArtistInviteConfigurableCell {
...
+        else {
+            stopTimer()
+        }
+    }
+
+    private var timer: Timer?
+
+    private func startTimer() {
+        guard timer == nil else { return }
+        timer = Timer.scheduledTimer(timeInterval: 0.25, target: self, selector: #selector(updateDateText), userInfo: nil, repeats: true)
```
alanpeabody: what are these interval units?

colinta: In seconds - I figured I wasn't burdening the processor much with having ¼-second precision.

